PR #1, opened 2024-05-21
Go library initial support

tudor-timcu:
Initial support for statically linking the go library into the C++ extension for PHP.
Features support for extracting the hostname from URLs extracted from hooking curl_init and curl_setopt from PHP.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
         phpize
         cd ../../build
         go build -buildmode=c-archive -o libaikido_go.a ../src/lib/aikido_lib.go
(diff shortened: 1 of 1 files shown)


src/extension/run-tests.php
```diff
@@ -0,0 +1,4192 @@
+#!/usr/bin/env php
```
willem-delbare: If this is a 3rd party lib and notnour code lets put it in a /vendor ?

tudor-timcu: It's a PHP script generated by the Zend framework when you generate the initial skeleton for the extension. For now it won't be used much. Sure, I can move it to /vendor.

tudor-timcu: Actually this is generated automatically in the build process, so I can remove it from the repo.

PR #2, opened 2024-05-23
Extended cpp and go support

tudor-timcu:
- Dynamic and easy way of hooking function in C++ with extended support for re-using common handlers
- Developed GO support for parsing JSON for C++ and storing and calling handlers for each function_execution type
- Specific handler for curl that extract, parse and store the outgoing hostnames
- Specific handler for various types of shell execution that extracts and stores the command line
- Reorganized GO code in multiple files

### DIFF
--- a/src/extension/aikido.cpp
+++ b/src/extension/aikido.cpp
@@ -28,8 +28,20 @@ struct FUNCTION_HANDLERS {
 	zif_handler original_handler;
 };
 
-#define AIKIDO_REGISTER_HANDLER(function_name) { std::string(#function_name), { handle_##function_name, nullptr } }
+/* 
+	Macro for registering an Aikido handler in the HOOKED_FUNCTIONS map.
+	It takes as parameters the PHP function name to be hooked and C++ function 
+		that should be called when that PHP function is executed.
+	The nullptr part is a placeholder where the original function handler from
+		the Zend framework will be stored at initialization when we run the hooking.
+*/
 #define AIKIDO_REGISTER_HANDLER_EX(function_name, function_pointer) { std::string(#function_name), { function_pointer, nullptr } }
+
+/*
+	Shorthand version of AIKIDO_REGISTER_HANDLER_EX that constructs automatically the C++ function to be called.
+	For example, if function name is curl_init this macro will store { "curl_init", { handle_curl_init, nullptr } }.
+*/
+#define AIKIDO_REGISTER_HANDLER(function_name) { std::string(#function_name), { handle_##function_name, nullptr } }
 
 
 unordered_map<std::string, FUNCTION_HANDLERS> HOOKED_FUNCTIONS = {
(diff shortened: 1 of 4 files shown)


src/extension/aikido.cpp
```diff
@@ -18,30 +18,39 @@
...
+
 
 struct FUNCTION_HANDLERS {
 	zif_handler aikido_handler;
 	zif_handler original_handler;
 };
 
-#define AIKIDO_REGISTER_HANDLER(function_name) { #function_name, { handle_##function_name, nullptr } }
+#define AIKIDO_REGISTER_HANDLER(function_name) { std::string(#function_name), { handle_##function_name, nullptr } }
+#define AIKIDO_REGISTER_HANDLER_EX(function_name, function_pointer) { std::string(#function_name), { function_pointer, nullptr } }
```
willem-delbare, line 38: what doet the ex mean? (lets comment perhaps?)

src/extension/aikido.cpp
```diff
@@ -109,6 +118,35 @@ ZEND_NAMED_FUNCTION(handle_curl_setopt) {
 	AIKIDO_HANDLER_END(curl_setopt);
 }
 
+ZEND_NAMED_FUNCTION(handle_shell_execution) {
```
willem-delbare, line 130: do we have mechanisms to split this up? otherwise this will become a really big file ?

src/lib/main.go
```diff
@@ -0,0 +1,34 @@
...
+	fmt.Println("[AIKIDO-GO] OnEvent:", eventJson)
+
+	var event map[string]interface{}
+	err := json.Unmarshal([]byte(eventJson), &event)
+	if err != nil {
+		log.Fatalf("Error parsing JSON: %s", err)
+	}
+
+	eventName := MustGetFromMap[string](event, "event")
+	data := MustGetFromMap[map[string]interface{}](event, "data")
```
willem-delbare, line 33: if it's called mustgetfrommap i'd almost expect it to throw an exeption if it's not in there. What happens if it's not in there?

src/lib/utils.go
```diff
@@ -0,0 +1,40 @@
+package main
+
+import (
+	"log"
+	"net/url"
+)
+
+func ExitIfKeyDoesNotExistInMap[K comparable, V any](m map[K]V, key K) {
```
willem-delbare: what does exit mean? kill the process?

src/extension/aikido.cpp
tudor-timcu: EX stands for Extended and in this case it's a different C++ macro that allows us to also set a custom function pointer to be called, for cases when we want to set the same function pointer to handle multiple hooked php functions (like the example for shell execution functions that all have the same field being the command line, so we can use just one handler to service 6 different php functions).

src/extension/aikido.cpp
tudor-timcu: Yes, I did not get to this part yet. For now we only have aikido.cpp and GoWrappers.cpp, but i'll split them up in the next PRs.

src/lib/main.go
tudor-timcu: It exists now, I'll change to throwing exceptions.

src/lib/utils.go
tudor-timcu: Yes.

PR #3, opened 2024-05-24
More functionality

tudor-timcu:
- Reorganized C++ sources
- Added support for generic hooking of object constructor and methods from C++ extension
- End to end flow of hooking PDO constructor and query function, with support in PHP extension and Go library


src/extension/HandlePDO.cpp
```diff
@@ -0,0 +1,73 @@
...
+		{ "event", "method_executed" },
+		{ "data", {
+			{ "class_name", "pdo" },
+			{ "method_name", "__construct" },
+			{ "parameters", {
+				{ "data_source", data_source_string }
+			} }
+		} }
+	};
+
```
willem-delbare, line 26: i don't think we'll ever need that password or username right? let's not capture this?

src/extension/Utils.cpp
```diff
@@ -0,0 +1,7 @@
+#include "Utils.h"
+
+std::string to_lowercase(const std::string& str) {
```
willem-delbare, line 3: when would we need this in c?

src/extension/HandlePDO.cpp
```diff
@@ -0,0 +1,63 @@
...
+			{ "method_name", "query" },
+			{ "parameters", {
+				{ "query",  query_string }
+			} }
+		} }
+	};
+
+	GoOnEvent(pdo_query);
+
+	if (strstr(query, "FROM users") != nullptr) {
```
willem-delbare: haha, best not commit :)

src/lib/handle_pdo.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+import "fmt"
+
+func OnMethodExecutedPdoConstruct(parameters map[string]interface{}) string {
+	data_source := MustGetFromMap[string](parameters, "data_source")
+	fmt.Println("[AIKIDO-GO] Got PDO data_source:", data_source)
```
willem-delbare: would still avoid printing this, i believe the soruce may sometimes contain passwords as well

willem-delbare: or at least sensitive stuff, let's not log it by default

PR #4, opened 2024-05-29
Generic handler for C++, interpret output event and cleanup

tudor-timcu:
- Added a generic wrapper handler for the C++ extensions that is called for any function we hook
- This handler performs all the necessary extraction of the execution context, than calls the specific handler for each function(curl_init, pdo_query, etc...)
- The specific handler returns an input event to be sent to the Go library
- The generic handler sends the input event to Go and gets an output event that is converted back to json
- The output event is parsed and interpreted, currently with 3 possible actions: block (the initial function is not executed anymore and the execution continues in the PHP script), throw (an exception is thrown to PHP with the parameters provided: message and code) or continue (that executes the initial handler and continues the normal PHP execution)

### DIFF
--- a/src/extension/Handle.cpp
+++ b/src/extension/Handle.cpp
@@ -27,10 +27,7 @@ enum ACTION {
 };
 
 ACTION aikido_execute_output(json event) {
-	if ( event["action"] == "block" ) {
-		return BLOCK;
-	}
-	else if ( event["action"] == "throw" ) {
+	if ( event["action"] == "throw" ) {
 		std::string message = event["message"].get<std::string>();
 		int code = event["code"].get<int>();
 		zend_throw_exception(zend_exception_get_default(), message.c_str(), code);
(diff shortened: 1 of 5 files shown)


src/extension/Handle.cpp
```diff
@@ -19,4 +19,84 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
...
+		zend_class_entry* executed_scope = zend_get_executed_scope();
+		
+		std::string function_name(ZSTR_VAL(func->common.function_name));
+		function_name = to_lowercase(function_name);
+
+		aikido_handler handler = nullptr;
+
+		std::string scope_name;
+
+		if (executed_scope) {
```
willem-delbare, line 54: can you comment what executed scope is? and why the if/else?

src/extension/HandleShellExecution.cpp
```diff
@@ -15,7 +13,7 @@ ZEND_NAMED_FUNCTION(handle_shell_execution) {
 
 	std::string functionNameString(AIKIDO_GET_FUNCTION_NAME());
 	
-	json shell_execution_event = {
+	event = {
```
willem-delbare: is this inputEvent in the other place , if so, let's align the names?

tudor-timcu: Yep, it's inputEvent, i'll update

src/extension/Handle.cpp
tudor-timcu: the executed_scope is null if it's a function and set to the class scope if a method is called

PR #5, opened 2024-05-29
Logging, config variables and E2E testing support

tudor-timcu:
- Custom configs loaded from php.ini and passed between C++ and GO to be used (log level, aikido token, blocking, ...)
- Log levels and overall logging support in both C++ and GO
- Extended testing support using .phpt files
- Updated pipeline to also test each change by running all cases in the test folder
- Use blocking mode variable to run output actions
- Sql injection POC with E2E flow
- Update pipeline to also build and install curl extension needed for tests

### DIFF
--- a/src/extension/Aikido.cpp
+++ b/src/extension/Aikido.cpp
@@ -41,9 +41,12 @@ static PHP_GINIT_FUNCTION(aikido)
 
 PHP_MINIT_FUNCTION(aikido)
 {
+	/* Register Aikido-specific (log level, blocking, token, ...) entries in php.ini */
 	REGISTER_INI_ENTRIES();
 
 	if (!aikido_global_init()) {
+		/* If the global initialization fails, we do not load the aikido extension */
+		/* The php script will still run, but without the aikido extension */
 		return FAILURE;
 	}
 
@@ -73,6 +76,7 @@ PHP_MINIT_FUNCTION(aikido)
 
 PHP_MSHUTDOWN_FUNCTION(aikido)
 {
+	/* Unregister Aikido-specific (log level, blocking, token, ...) entries in php.ini */
 	UNREGISTER_INI_ENTRIES();
 	return SUCCESS;
 }


src/lib/utils.go
```diff
@@ -36,5 +37,5 @@ func GetDomain(rawurl string) string {
 	if err != nil {
 		return ""
 	}
-	return parsedURL.Hostname()
+	return strings.TrimPrefix(parsedURL.Hostname(), "www.")
```
willem-delbare: i think we do need full subdomains here!

src/extension/Aikido.cpp
```diff
@@ -3,18 +3,57 @@
...
+	if (AIKIDO_GLOBAL(log_level) < AIKIDO_LOG_LEVEL_DEBUG ||
+		AIKIDO_GLOBAL(log_level) > AIKIDO_LOG_LEVEL_ERROR) {
+		AIKIDO_GLOBAL(log_level) = AIKIDO_LOG_LEVEL_ERROR;
+	}
+
+	const char* log_level_str = aikido_log_level_str((AIKIDO_LOG_LEVEL)AIKIDO_GLOBAL(log_level));
+	
+	AIKIDO_LOG_DEBUG("Config:\n");
+	AIKIDO_LOG_DEBUG("Log level: %s\n", log_level_str);
+	AIKIDO_LOG_DEBUG("Token: %s\n", AIKIDO_GLOBAL(token));
```
willem-delbare: best never log the token right?

src/extension/Aikido.cpp
```diff
@@ -3,18 +3,56 @@
...
 	ZEND_TSRMLS_CACHE_UPDATE();
 #endif
+}
+
+PHP_MINIT_FUNCTION(aikido)
+{
+	REGISTER_INI_ENTRIES();
+
+	if (!aikido_global_init()) {
+		return FAILURE;
```
willem-delbare, line 50: Why would this ever fail? And would this lead to app crash, the return fail?
Lets add comments

src/extension/Aikido.cpp
```diff
@@ -35,6 +73,7 @@ PHP_MINIT_FUNCTION(aikido)
 
 PHP_MSHUTDOWN_FUNCTION(aikido)
 {
+	UNREGISTER_INI_ENTRIES();
```
willem-delbare, line 80: Lets add commetns what this does. It doesnt actually remove stuff from php.ini? Or does it?

PR #6, opened 2024-06-03
More functionality

tudor-timcu:
- Prepared polling "thread" code for pulling the cloud config
- Added GoUninit to kill the goroutine when the module is unloaded
- Refactor & git ignore
- Added support for setting the cloud endpoint
- Support for overriding the php.ini aikido config with env variables (for token and cloud endpoint)
- Fix issues that were causing PHP-7.3 & PHP-7.4 to not build
- Updated tests

### DIFF
--- a/tests/test_curl.phpt
+++ b/tests/test_curl.phpt
@@ -4,7 +4,7 @@ Test curl initialization and setting options
 --INI--
 extension=curl.so
 extension=aikido.so
-aikido.log_level=1
+aikido.log_level=2
 
 --SKIPIF--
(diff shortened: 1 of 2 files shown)


src/lib/cloud/cloud.go
```diff
@@ -0,0 +1,69 @@
...
+	close(stop)
+	wg.Wait()
+}
+
+func Init(configuredEndpoint string) {
+	endpoint = configuredEndpoint
+	// StartConfigThread()
+}
+
+func Uninit() {
```
willem-delbare, line 67: comments

PR #7, opened 2024-06-04
Support for sending Started event to cloud

tudor-timcu:
- Added support for Started event
- Added support in Utils for: hostname, os info and ip address extraction
- Cleaner json unmarshalling on init
- Types package to group all the types, including the ones for Started event
- Globals package to store global init data and constants
- Abstract cloud requests via SendEvent function
- Utils package to store utility functions

### DIFF
--- a/src/lib/aikido_types/handle.go
+++ b/src/lib/aikido_types/handle.go
@@ -1,10 +1,6 @@
 package aikido_types
 
-type EventFunctionExecutedFn func(map[string]interface{}) string
-
-type FunctionExecutedHandlersFn func(map[string]interface{}) string
-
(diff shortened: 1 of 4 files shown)


src/lib/aikido_types/handle.go
```diff
@@ -0,0 +1,12 @@
+package aikido_types
+
+type EventFunctionExecutedFn func(map[string]interface{}) string
+
+type FunctionExecutedHandlersFn func(map[string]interface{}) string
+
+type MethodExecutedHandlersFn func(map[string]interface{}) string
+
```
willem-delbare: These 3 all kind of sound/look alike. Can we add comments or have more unique names?

PR #8, opened 2024-06-04
Extended build support

tudor-timcu:
- Added support for building, testing and generating artifacts for PHP versions from 7.3 to 8.4
- Get aikido version from php_aikido.h and use it to version artifacts
- Use setup-php github action to reduce build & testing to ~1 minute per version
- Store test failures in case of failure for debugging

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
         echo "AIKIDO_ARTIFACT=aikido-$AIKIDO_VERSION-php-${{ matrix.php_version }}" >> $GITHUB_ENV
 
     - name: Setup PHP
-      uses: AikidoSec/setup-php@2
+      uses: AikidoSec/setup-php@v2
       with:
         php-version: ${{ matrix.php_version }}
         extensions: :curl


.github/workflows/build.yml
```diff
@@ -27,34 +25,25 @@ jobs:
...
-        cd php-src
-        ./buildconf
-        ./configure --prefix=$HOME/php --with-config-file-path=$HOME/php
-        make -j$(nproc)
-        make install
-
-    - name: Set PHP PATH
-      run: echo "PATH=$HOME/php/bin:$PATH" >> $GITHUB_ENV
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
```
willem-delbare: https://github.com/AikidoSec/setup-php

.github/workflows/build.yml
```diff
@@ -66,17 +55,27 @@ jobs:
...
-    - name: Test
+    - name: Run Aikido tests
       run: |
         cd ${{ github.workspace }}
-        php src/extension/run-tests.php ./tests
+        export TEST_PHP_EXECUTABLE=/usr/bin/php
+        /usr/bin/php src/extension/run-tests.php ./tests
+
+    - name: Delete existing artifact
+      uses: geekyeggo/delete-artifact@v5
```
willem-delbare: Seems dangerous to delete old artifacts? Dont think its needed?

PR #22, opened 2024-07-24
Improvements on code & tests & pipeline

tudor-timcu:
- Debian package build (dpkg) & tests
- CLI automated tests for Ubuntu 20.04
- CLI automated tests for Ubuntu 22.04
- CLI automated tests for CentOS 8
- Automatic check for dependencies on the RPM / DEB package (check to not use too new versions of GLIBC / GLIBCXX)
- Graceful handling and logging for exception cases
- Handle both CLI & FPM sapi modes (on CLI do not load agent, only extension + request processor - used for tests and playing with the CLI)
- Config only loaded from ENV variables (dropped config.json from disk)

### DIFF
--- a/include/Utils.h
+++ b/include/Utils.h
@@ -38,12 +38,12 @@ void aikido_log_uninit();
 void aikido_log(AIKIDO_LOG_LEVEL level, const char* format, ...);
 
 
-//#if defined(ZEND_DEBUG)
+#if defined(ZEND_DEBUG)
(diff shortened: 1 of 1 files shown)


include/Utils.h
```diff
@@ -38,29 +38,25 @@ void aikido_log_uninit();
 void aikido_log(AIKIDO_LOG_LEVEL level, const char* format, ...);
 
 
-#if defined(ZEND_DEBUG)
+//#if defined(ZEND_DEBUG)
```
willem-delbare: prob not on purpose?

PR #24, opened 2024-07-29
API endpoints reporting and other minor improvements

tudor-timcu:
- Report API endpoints with Route & Methods & Hits
- Handle graceful exit when PHP-FPM does not run as root (more generically for cases when the unix socket cannot be created or gRPC server cannot be initialized)
- Improved filename logging for better tracking

### DIFF
--- a/lib/request-processor/grpc/client.go
+++ b/lib/request-processor/grpc/client.go
@@ -37,6 +37,7 @@ func Uninit() {
 	conn.Close()
 }
 
+/* Send outgoing domain to Aikido Agent via gRPC */
 func OnReceiveDomain(domain string) {
 	if client == nil {
 		return
@@ -53,6 +54,7 @@ func OnReceiveDomain(domain string) {
 	log.Debugf("Domain sent via socket: %v", domain)
 }
 
+/* Send request metadata (route & method) to Aikido Agent via gRPC */
 func OnReceiveRequestMetadata(method string, route string) {
 	if client == nil {
 		return


include/Utils.h
```diff
@@ -60,3 +60,6 @@ std::string get_environment_variable(const std::string& env_key);
 std::string config_override_with_env(const std::string& env_key, const std::string default_value);
 
 bool config_override_with_env_bool(const std::string& env_key, bool default_value);
+
+bool send_request_data_to_agent();
```
willem-delbare: this only shares domain-level data right?
let's rename 'share_outbound_domain_data..'

lib/agent/grpc/server.go
```diff
@@ -26,7 +26,36 @@ func (s *server) OnReceiveDomain(ctx context.Context, req *protos.Domain) (*empt
 	return &emptypb.Empty{}, nil
 }
 
-func Init() {
+func (s *server) OnReceiveHttpRequestInfo(ctx context.Context, req *protos.HttpRequestInfo) (*emptypb.Empty, error) {
```
willem-delbare: lets clarify in a comment: weonly recieve method + routes here!

lib/php-extension/Utils.cpp
```diff
@@ -86,3 +90,55 @@ bool config_override_with_env_bool(const std::string& env_key, bool default_valu
...
+        route = route.substr(0, pos);
+    }
+    json result = {
+        {"route", route},
+        {"method", method}
+    };
+    return result;
+}
+
+bool send_request_data_to_agent(){
```
willem-delbare: let's call it metadata , we're never sending body etc

lib/php-extension/Aikido.cpp
```diff
@@ -182,7 +182,14 @@ PHP_RINIT_FUNCTION(aikido) {
 
 		AIKIDO_LOG_DEBUG("Aikido Request Processor initialized successfully!\n");
 	}
-
+	
+	zend_string *server_str = zend_string_init("_SERVER", sizeof("_SERVER") - 1, 0);
+	if (server_str){
+		zend_is_auto_global(server_str); 
+		zend_string_release(server_str);
```
willem-delbare: can you comment what this code does?

lib/request-processor/grpc/client.go
```diff
@@ -52,3 +52,19 @@ func OnReceiveDomain(domain string) {
 
 	log.Debugf("Domain sent via socket: %v", domain)
 }
+
+func OnReceiveHttpRequestInfo(method string, route string) {
```
willem-delbare: metadata again? the client is the one sending it not? Shouldn't this function be named send?
if that's not possible, let's commend to clarify things

lib/request-processor/handle_http_request_info.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"main/grpc"
+	"main/log"
+	"main/utils"
+)
+
+func OnHttpRequestInfo(data map[string]interface{}) string {
```
willem-delbare: meta..

PR #27, opened 2024-08-04
Rate limiting support

tudor-timcu:
- Rate limiting support
- Passing endpoint configs to request processors (polling thread that gets config once every minute)
- Checking if a route + method is rate limited and if so do a sync call via gRPC with timeout
- If it's not rate limited, do async call as before
- In agent, for routes that are rate limited, store buckets of 1 minute in a queue that acts as a sliding window and keeps the moving total for the past x minutes that are configured for rate limiting
- Use the stored data to reply to the blocking requests with the status of rate limiting, if the request should be forwarded to server or not
- If it shouldn't be forwarded to server (rate limited), throw exception to the php server layer
- Generic support for PHP templates to be executed in the C++ layer, via a custom action called "exit" that evaluates PHP code that sets http response code + message + exit
- Handling rate limiting scenarios by using this new output action
- Some refactoring

### DIFF
--- a/include/Includes.h
+++ b/include/Includes.h
@@ -22,6 +22,7 @@
 #include "ext/standard/info.h"
 #include "GoCGO.h"
 #include "php_aikido.h"
+#include "PhpTemplates.h"
 
 #include "3rdparty/json.hpp"
 
--- a/include/PhpTemplates.h
+++ b/include/PhpTemplates.h
@@ -1,0 +1,6 @@
+#pragma once
+
+#define PHP_EXIT_ACTION_TEMPLATE "http_response_code(%d);\n" \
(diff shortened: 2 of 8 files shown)


lib/request-processor/grpc/config.go
```diff
@@ -0,0 +1,69 @@
...
+	"main/globals"
+	"main/ipc/protos"
+	"time"
+)
+
+var (
+	stop              chan struct{}
+	cloudConfigTicker = time.NewTicker(1 * time.Minute)
+)
+
```
willem-delbare, line 14: could you comment the goal of this class? It's to transmit a (partial) copy of the config to the child threads?

PR #30, opened 2024-08-09
Path traversal detection

tudor-timcu:
- Path traversal detection
- Support for extracting url, https, method, query, header, remote address & cookies - all necessary for detection
- Support for extracting and parsing the body
- Pass data on a new event called "path_accessed" to the request processor for checking the detection (along with the request context)
- Hook multiple PHP functions that do path operations (basename, chgrp, chmod, chown, clearstatcache, copy, dirname, disk_free_space), extract path parameter and pass to the "path_accessed" event
- Support for functions that have either one or two path arguments
- Added test for all the new functionality
- Added support for ExtractStringsFromUserInput, ParseFormData and tryDecodeAsJWT

### DIFF
--- a/lib/php-extension/Handle.cpp
+++ b/lib/php-extension/Handle.cpp
@@ -17,6 +17,7 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
 	
 	/* Path traversal */
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(basename,             handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chdir,             	  handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chgrp,	              handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chmod,                handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chown,                handle_file_path_access),
@@ -55,6 +56,7 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(lstat,                handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(mkdir,                handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(move_uploaded_file,   handle_file_path_access_2),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(opendir,              handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(parse_ini_file,       handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(pathinfo,             handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(readfile,             handle_file_path_access),
@@ -62,6 +64,7 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(realpath,             handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(rename,               handle_file_path_access_2),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(rmdir,                handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(scandir,              handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(stat,                 handle_file_path_access),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(symlink,              handle_file_path_access_2),
 	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(touch,                handle_file_path_access),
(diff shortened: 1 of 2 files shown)


lib/php-extension/Handle.cpp
```diff
@@ -1,18 +1,72 @@
...
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(pathinfo,             handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(readfile,             handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(readlink,             handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(realpath,             handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(rename,               handle_file_path_access_2),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(rmdir,                handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(stat,                 handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(symlink,              handle_file_path_access_2),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(touch,                handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(unlink,               handle_file_path_access),
```
hansott, line 71: * https://www.php.net/manual/en/function.scandir.php
* https://www.php.net/manual/en/function.chdir.php
* https://www.php.net/manual/en/function.opendir.php

lib/php-extension/HandlePathTraversal.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    std::string functionNameString(AIKIDO_GET_FUNCTION_NAME());
+    
+    inputEvent = {
+        { "event", "function_executed" },
+        { "data", {
+            { "function_name", "path_accessed" },
+            { "parameters", {
+                { "filename", filenameString },
+                { "filename2", "" },
```
hansott: maybe `NULL` is safer here? To be explicit that there isn't a path?

lib/php-extension/Utils.cpp
```diff
@@ -115,11 +114,67 @@ json get_route_and_method(zval *server) {
...
+    long maxlen = PHP_STREAM_COPY_ALL;
+    zend_string *contents;
+    php_stream *stream;
+
+    stream = php_stream_open_wrapper("php://input", "rb", 0 | REPORT_ERRORS, NULL);
+    if ((contents = php_stream_copy_to_mem(stream, maxlen, 0)) != NULL) {
+        php_stream_close(stream);
+        return std::string(ZSTR_VAL(contents));
+    }
+    php_stream_close(stream);
```
hansott, line 134: Do we need to rewind the stream so that user code can read it like normal?

lib/php-extension/Utils.cpp
```diff
@@ -115,11 +114,67 @@ json get_route_and_method(zval *server) {
...
+    bool https = extract_server_var(server, "HTTPS") != "" ? true : false;
+
+    if (server && Z_TYPE_P(server) == IS_ARRAY) {
+        return {
+            { "https", https },
+            { "url", (https ? "https://" : "http://") + extract_server_var(server, "HTTP_HOST") + extract_server_var(server, "REQUEST_URI") },
+            { "method", extract_server_var(server, "REQUEST_METHOD") },
+            { "query", extract_server_var(server, "QUERY_STRING") },
+            { "headers", extract_headers(server) },
+            { "remoteAddress", extract_server_var(server, "REMOTE_ADDR") },
```
hansott, line 169: Should we check the X-Forwarded-For header? https://github.com/AikidoSec/firewall-node/blob/main/library/helpers/getIPAddressFromRequest.ts#L3

e.g. when behind a load balancer

lib/request-processor/vulnerabilities/path-traversal/checkContextForPathTraversal.go
```diff
@@ -0,0 +1,36 @@
+package path_traversal
+
+import (
+	"main/helpers"
+	"main/utils"
+)
+
+var SOURCES = []string{"body", "routeParams", "query", "params", "headers", "cookies"}
```
hansott, line 8: could be handy to move this to a separate file in the future

lib/request-processor/vulnerabilities/path-traversal/checkContextForPathTraversal.go
```diff
@@ -0,0 +1,36 @@
...
+
+type Context map[string]interface{}
+
+func CheckContextForPathTraversal(filename string, operartion string, context Context, checkPathStart bool) *utils.InterceptorResult {
```
hansott, line 12: checkPathStart is never False?

hansott: I think we added that as a perf optimization in node

lib/php-extension/HandlePathTraversal.cpp
tudor-timcu: nlohmann json library does not support null fields. I'll remove it.

lib/php-extension/Utils.cpp
tudor-timcu: It's just a copy of the stream, does not affect the original one.

lib/php-extension/Utils.cpp
tudor-timcu: I'll do a separate PR with this.

lib/request-processor/vulnerabilities/path-traversal/checkContextForPathTraversal.go
tudor-timcu: I'll do it in a next PR.

PR #37, opened 2024-08-21
Allowed ip addresses support

tudor-timcu:
- Added support for allowed ip addresses for specific routes (a route configured to only allow access from specific IPs)
- Added support for IPs that are excluded from rate limiting


lib/agent/grpc/request.go
```diff
@@ -40,6 +41,13 @@ func updateRateLimitingStatus(req *protos.RequestMetadataShutdown) {
 	rateLimitingData.Status.NumberOfRequestsPerWindow.IncrementLast()
 }
 
+func isIpAllowedForRateLimiting(ip string) bool {
```
willem-delbare: do you mean the bypass list for ip's that are never rate limiting? let's rename it a bit for clarity

willem-delbare: or 'shouldIPBeRatelimited etc

PR #44, opened 2024-08-30
Automated E2E tests

tudor-timcu:
- Added E2E testing framework + mock for aikido core
- Added execution of tests in the github pipeline for 'ubuntu-latest' & 'ubuntu-20.04' (centos will follow later)
- Added testing scenarios for: allowed ips on route, bypassed ips from rate limiting, user reporting, user blocking, domains reporting, rate limiting for 1 minute, rate limiting for 10 minutes, received any stats, ... (more to follow)
- For now these tests run using PHP simple server (php -S ...). In the next iterations I'll add support for running the exact same tests with nginx & apache setups.
- Added env variable support to control if the localhost IP should be bypassed by default (it is useful in tests to be able to use the localhost ip address to test for rate limiting, detections, ...). The env variable is True by default, and in tests is set to False.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -315,21 +315,16 @@ jobs:
           export TEST_PHP_EXECUTABLE=/usr/bin/php
           php lib/php-extension/run-tests.php ./tests/cli
 
-      - name: Run simple server tests
-        run: |
-          ls -l /opt/aikido-${{ env.AIKIDO_VERSION }}/
-          ls -l /var/log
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -312,16 +315,28 @@ jobs:
           export TEST_PHP_EXECUTABLE=/usr/bin/php
           php lib/php-extension/run-tests.php ./tests/cli
 
+      #- name: Run simple server tests
```
willem-delbare, line 318: why is it commented?

PR #45, opened 2024-09-02
Added support for parallel tests execution

tudor-timcu:
- Generate socket path at init in C++ extension and pass socket path to request-processor and agent
- Enable multithreading support for tests for parallel execution
- Enable E2E tests on CentOs
- Improved aikido rpm (aikido.spec) to group all socket files in /run/aikido-{version} and disable using of 'sudo' as it's incompatible in dockers and rpm commands run by default with sudo

### DIFF
--- a/lib/request-processor/aikido_types/config.go
+++ b/lib/request-processor/aikido_types/config.go
@@ -1,11 +1,11 @@
 package aikido_types
 
 type EnvironmentConfigData struct {
-	LogLevel                  string `json:"log_level"`
-	SocketPath                string `json:"socket_path"`
-	SAPI                      string `json:"sapi"`
-	TrustProxy                bool   `json:"trust_proxy"`
-	LocalhostAllowedByDefault bool   `json:"localhost_allowed_by_default"`
+	SocketPath                string `json:"socket_path"`                  // '/run/aikido-{version}/aikido-{datetime}-{randint}.sock'
+	LogLevel                  string `json:"log_level"`                    // default: 'INFO'
+	SAPI                      string `json:"sapi"`                         // '{php-sapi}'
+	TrustProxy                bool   `json:"trust_proxy"`                  // default: true
+	LocalhostAllowedByDefault bool   `json:"localhost_allowed_by_default"` // default: true
 }
 
 type RateLimiting struct {
--- a/lib/request-processor/utils/utils.go
+++ b/lib/request-processor/utils/utils.go
@@ -117,14 +117,35 @@ func isLocalhost(ip string) bool {
 }
 
 func IsIpAllowed(allowedIps map[string]bool, ip string) bool {
-	return (globals.EnvironmentConfig.LocalhostAllowedByDefault && isLocalhost(ip)) || len(allowedIps) == 0 || KeyExists(allowedIps, ip)
+	if globals.EnvironmentConfig.LocalhostAllowedByDefault && isLocalhost(ip) {
+		return true
+	}
+
+	if len(allowedIps) == 0 {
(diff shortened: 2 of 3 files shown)


lib/php-extension/Utils.cpp
```diff
@@ -325,3 +325,19 @@ bool aikido_call_user_function_one_param(std::string function_name, std::string
...
+    std::random_device rd;
+    std::mt19937 gen(rd());
+    std::uniform_int_distribution<> dis(100000, 999999);
+    return dis(gen);
+}
+
+std::string aikido_get_socket_path() {
+    std::time_t current_time = std::time(nullptr);
+    char time_str[20];
+    std::strftime(time_str, sizeof(time_str), "%Y%m%d%H%M%S", std::localtime(&current_time));
```
willem-delbare, line 339: what's the logic behind this? you'll get a different answer if you call this again later? It's only supposed to be called once? in that case let's call it generateRandomSocketPathName ?

lib/request-processor/aikido_types/config.go
```diff
@@ -1,9 +1,11 @@
...
-type InitConfigData struct {
-	LogLevel   string `json:"log_level"`
-	SAPI       string `json:"sapi"`
-	TrustProxy bool   `json:"trust_proxy"`
+type EnvironmentConfigData struct {
+	LogLevel                  string `json:"log_level"`
+	SocketPath                string `json:"socket_path"`
+	SAPI                      string `json:"sapi"`
+	TrustProxy                bool   `json:"trust_proxy"`
+	LocalhostAllowedByDefault bool   `json:"localhost_allowed_by_default"`
```
willem-delbare: can we add in comments the deafult states of these? hard to find

lib/request-processor/utils/utils.go
```diff
@@ -117,14 +117,14 @@ func isLocalhost(ip string) bool {
 }
 
 func IsIpAllowed(allowedIps map[string]bool, ip string) bool {
-	return isLocalhost(ip) || len(allowedIps) == 0 || KeyExists(allowedIps, ip)
+	return (globals.EnvironmentConfig.LocalhostAllowedByDefault && isLocalhost(ip)) || len(allowedIps) == 0 || KeyExists(allowedIps, ip)
```
willem-delbare: it's probably best to rewrite this with early returns instead of || &&

lib/request-processor/utils/utils.go
```diff
@@ -117,14 +117,14 @@ func isLocalhost(ip string) bool {
...
 }
 
 func IsIpBypassed(ip string) bool {
 	globals.CloudConfigMutex.Lock()
 	defer globals.CloudConfigMutex.Unlock()
 
-	return isLocalhost(ip) || KeyExists(globals.CloudConfig.BypassedIps, ip)
+	return (globals.EnvironmentConfig.LocalhostAllowedByDefault && isLocalhost(ip)) || KeyExists(globals.CloudConfig.BypassedIps, ip)
```
willem-delbare: it's probably best to rewrite this with early returns instead of || &&

package/rpm/aikido.spec
```diff
@@ -71,21 +71,24 @@ else
...
-        echo "Failed to remove the socket."
+        echo "Failed to remove the socket folder."
     fi
 else
-    echo "Socket $SOCKET_PATH does not exist."
+    echo "Socket $SOCKET_FOLDER does not exist."
 fi
 
+mkdir -p /run/aikido-%{version}
+chmod 777 /run/aikido-%{version}
```
willem-delbare, line 90: is this permitted usually?

PR #47, opened 2024-09-03
Added support for detection event + caching context data

tudor-timcu:
- Added support for callbacks between Go & C++ layers. This allows the Go code to request ONLY when needed the data from the PHP layer (like body, query, route, …).
- The data obtained from these callbacks is cached on the Go layer for future use. The mechanism used is a “lazy” type of cache, that when some data is requested, if it wasn’t already stored, that it is requested from the C++ layer (parsed if necessary) and that stored in the cache. Subsequent calls returns the computed values directly without additional processing or memory copies.
- The cached context data is now used everywhere in the request-processor when it’s needed (route, method, body, query, headers, …)
- Simplified API between C++ & Go as now I don’t need to pass context to any event that might need it -> the Go layer will just request via callback what data it needs, when it needs it.
- Moved include folder into lib/php-extesion as it’s only C++ related
- Added support for detected_attack event. On detection, data is passed asynchronous from the request processor to agent via gRPC and the agent makes the cloud requests.
- Disabled some unit tests that were testing with the context passed as parameter (need to update those to load the context into cache - will do later)
- Added new E2E test for checking the path traversal blocking on attack

### DIFF
--- a/lib/request-processor/context/cache.go
+++ b/lib/request-processor/context/cache.go
@@ -19,7 +19,7 @@ import (
 
 type ParseFunction func(string) map[string]interface{}
 
-func ContextCacheMap(contextId int, rawDataPtr **string, parsedPtr **map[string]interface{}, stringsPtr **map[string]string, parseFunc ParseFunction) {
+func ContextSetMap(contextId int, rawDataPtr **string, parsedPtr **map[string]interface{}, stringsPtr **map[string]string, parseFunc ParseFunction) {
 	if *stringsPtr != nil {
 		return
 	}
@@ -35,7 +35,7 @@ func ContextCacheMap(contextId int, rawDataPtr **string, parsedPtr **map[string]
 	*stringsPtr = &strings
 }
 
-func ContextCacheString(context_id int, m **string) {
+func ContextSetString(context_id int, m **string) {
 	if *m != nil {
 		return
 	}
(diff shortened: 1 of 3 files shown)


lib/php-extension/GoWrappers.cpp
```diff
@@ -27,6 +27,81 @@ json GoRequestProcessorOnEvent(json& event) {
     return output;
 }
 
+char* GoContextCallback(int context_id) {
```
willem-delbare, line 33: can you comment when this function is used/intent?

lib/request-processor/context/cache.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"main/helpers"
+	"main/log"
+	"main/utils"
+	"strconv"
+)
+
+type ParseFunction func(string) map[string]interface{}
+
+func ContextCacheMap(contextId int, rawDataPtr **string, parsedPtr **map[string]interface{}, stringsPtr **map[string]string, parseFunc ParseFunction) {
```
willem-delbare: can you comment on how long this cache lives? this is in-process?

lib/request-processor/handle_user_event.go
```diff
@@ -26,5 +25,6 @@ func OnUserEvent(data map[string]interface{}) string {
 		return `{"action": "exit", "message": "You are blocked by Aikido Firewall!", "response_code": 403}`
 	}
 
+	context.ContextCacheUserId(id)
```
willem-delbare: not super clear this is a setter function

lib/request-processor/main.go
```diff
@@ -41,6 +44,37 @@ func RequestProcessorInit(initJson string) (initOk bool) {
...
+	}
+
+	contextData := C.call(CContextCallback, C.int(contextId))
+	if contextData == nil {
+		return ""
+	}
+
+	goContextData := C.GoString(contextData)
+
+	C.free(unsafe.Pointer(contextData))
```
willem-delbare: why is this needed? If unavoidaable, let's comment

lib/request-processor/attack/attack.go
```diff
@@ -0,0 +1,51 @@
+package attack
+
+import (
+	"main/context"
+	"main/ipc/protos"
+	"main/utils"
+)
+
+func GetMetadataProto(metadata map[string]string) []*protos.Metadata {
```
willem-delbare, line 10: let's comment what the goal of this fn is

PR #48, opened 2024-09-04
Multiple small improvements

tudor-timcu:
- AIK-3459: max <x> attacks per timeframe support
- AIK-3372: improve exception message by adding metadata
- AIK-3485: send platform name and version (ex: fpm-fcgi, php8.1)
- AIK-3461: namespace the \aikido\set_user function
- Improve flaky tests


tests/cli/path_traversal/test_path_traversal_touch.phpt
```diff
@@ -17,4 +17,4 @@ touch($file);
 ?>
 
 --EXPECTREGEX--
-.*Fatal error: Uncaught Exception: Path traversal detected.*
+.*Fatal error: Uncaught Exception: Aikido firewall has blocked a path traversal attack.*
```
willem-delbare, line 20: i don't think we want to give the attacker information about the kind of attack

lib/request-processor/attack/attack.go
```diff
@@ -52,3 +53,15 @@ func GetAttackDetectedProto(res utils.InterceptorResult) protos.AttackDetected {
 		},
 	}
 }
+
+func BuildAttackDetectedMessage(result utils.InterceptorResult) string {
```
willem-delbare, line 57: does this go to html output ? or why does it need escaping?

PR #51, opened 2024-09-09
Automated benchmarks & valgrind tests & performance improvements

tudor-timcu:
- Reworked C++ & Go communication to not rely on json as the back and forth json conversions were incurring performance penalty
- Cached both request data and event data in C++ and pass data to golang only via callback to pass native strings to increase performance
- Compile the release version for C++ binary and Go libs & strip binaries of debug symbols
- Fixed a crash after running with release version in PHP function parameter extraction
- Benchmark tests per request (benchmark/server tests)
- Benchmark tests per hooked PHP function (benchmark/cli tests)
- Valgrind tests

### DIFF
--- a/lib/php-extension/Handle.cpp
+++ b/lib/php-extension/Handle.cpp
@@ -138,6 +138,13 @@ ZEND_NAMED_FUNCTION(aikido_generic_handler) {
 			AIKIDO_LOG_DEBUG("Calling handler for \"%s\"!\n", scope_name.c_str());
 
 			EVENT_ID eventId = NO_EVENT_ID;
+			/*
+				The handler for a specific PHP function that we hook can set an event ID
+				to be sent to the Go libary (request processor).
+				This will notify the Go library that an event has happend in the PHP layer.
+				The event ID is initialy empty and it's only sent to Go only if the C++ handler
+				for the currently hooked function sets it.
+			*/
 			handler(INTERNAL_FUNCTION_PARAM_PASSTHRU, eventId);
 
 			if (eventId != NO_EVENT_ID) {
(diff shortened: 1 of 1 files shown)


lib/request-processor/main.go
```diff
@@ -80,35 +79,21 @@ func RequestProcessorContextInit(contextCallback C.ContextCallback) (initOk bool
 }
 
 //export RequestProcessorOnEvent
-func RequestProcessorOnEvent(eventJson string) (outputJson *C.char) {
+func RequestProcessorOnEvent(eventId int) (outputJson *C.char) {
```
willem-delbare, line 82: Is this eventueel freed?

lib/request-processor/handle_request_metadata.go
```diff
@@ -52,24 +56,23 @@ func OnRequestInit(data map[string]interface{}) string {
...
-		return "{}"
+		return
 	}
 
 	go grpc.OnRequestShutdown(method, route, status_code, 10*time.Millisecond)
+}
 
-	return "{}"
+func OnRequestShutdown() string {
+	go OnRequestShutdownReporting(context.GetMethod(), context.GetRoute(), context.GetStatusCode())
```
willem-delbare, line 76: Why is this reported?

lib/php-extension/Handle.cpp
```diff
@@ -148,14 +155,16 @@ ZEND_NAMED_FUNCTION(aikido_generic_handler) {
...
+
 	if (original_handler) {
 		original_handler(INTERNAL_FUNCTION_PARAM_PASSTHRU);
 
 		if (!caughtException && post_handler) {
-			post_handler(INTERNAL_FUNCTION_PARAM_PASSTHRU, inputEvent);
-			if (!inputEvent.empty()) {
-				GoRequestProcessorOnEvent(inputEvent);
+			EVENT_ID eventId = NO_EVENT_ID;
+			post_handler(INTERNAL_FUNCTION_PARAM_PASSTHRU, eventId);
```
willem-delbare, line 178: Comments here to express the goal here would be good

PR #54, opened 2024-09-23
Sql injection detection

tudor-timcu:
- Support for SQL injection detection
- Golang unit tests
- Cli test
- E2E test
- Benchmark
- Only PDO support for now, more drivers to follow in a separate PR


lib/php-extension/HandlePDO.cpp
```diff
@@ -1,50 +1,36 @@
...
-				{ "query",  query_string }
-			} }
-		} }
-	};
-	*/
+	if (!query) {
+		return;
+	}
+
+	zval *pdo_object = getThis();
```
willem-delbare, line 21: what is get this? lets comment

PR #56, opened 2024-09-25
SSRF detection

tudor-timcu:
- Added SSRF detection support
- Added unittests for golang SSRF code
- Added tests (cli & E2E) and benchmarks for SSRF
- Added support to treat file access functions that have an URL passed as parameter as outgoing requests (eg: file_get_contents can take both a file path and an URL as parameter)
- Apply SSRF checks for file open functions that can take an URL (file_get_contents, fopen, ...) + added support for these functions to report outgoing hostnames to the agent
- Check if domains resolve to private IPs for SSRF TOCTOU
- Check IMDS ips
- Updated readme

### DIFF
--- a/lib/request-processor/context/event_getters.go
+++ b/lib/request-processor/context/event_getters.go
@@ -47,7 +47,7 @@ func GetModule() string {
 	return Context.Callback(C.MODULE)
 }
 
-func getHostNameAndPort(urlCallbackId int) (string, int) {
+func getHostNameAndPort(urlCallbackId int) (string, int) { // urlcallbackid is the type of data we request, eg C.OUTGOING_REQUEST_URL
 	urlStr := Context.Callback(urlCallbackId)
 	urlParsed, err := url.Parse(urlStr)
 	if err != nil {


lib/API.h
```diff
@@ -37,7 +38,9 @@ enum CALLBACK_ID {
     FUNCTION_NAME,
 
     OUTGOING_REQUEST_URL,
+    OUTGOING_REQUEST_EFFECTIVE_URL,
```
willem-delbare: can you comment what effective URL is? Eg is it normalized? Or is it the url after redirects have resolved?

lib/php-extension/Handle.cpp
```diff
@@ -19,30 +19,32 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
...
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(touch, handle_file_path_access),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(unlink, handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chdir, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chgrp, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chmod, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chown, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(copy, handle_pre_file_path_access_2, handle_post_file_path_access_2),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file, handle_pre_file_path_access, handle_post_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file_size, handle_pre_file_path_access, handle_post_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file_exists, handle_pre_file_path_access, handle_post_file_path_access),
```
willem-delbare: doesn't seem related to SSRF, and i thought we removed file_exists and others?

lib/php-extension/Handle.cpp
```diff
@@ -19,30 +19,32 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
...
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(symlink, handle_file_path_access_2),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(touch, handle_file_path_access),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(unlink, handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chdir, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chgrp, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chmod, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chown, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(copy, handle_pre_file_path_access_2, handle_post_file_path_access_2),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file, handle_pre_file_path_access, handle_post_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file_size, handle_pre_file_path_access, handle_post_file_path_access),
```
willem-delbare: don't think we need this sink

lib/request-processor/context/cache.go
```diff
@@ -123,3 +123,7 @@ func ContextSetUserId() {
 func ContextSetUserName() {
 	ContextSetString(C.CONTEXT_USER_NAME, &Context.UserName)
 }
+
+func ContextSetPartialInterceptorResult(interceptorResult utils.InterceptorResult) {
```
willem-delbare: comment what a partial intercept result is

lib/request-processor/helpers/getHostnameAndPortFromURL.go
```diff
@@ -0,0 +1,35 @@
...
+		return 0
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return 0
+	}
+	return port
+}
+
+func FixURL(url string) string {
```
willem-delbare, line 24: lets comment why we do this

lib/php-extension/include/Utils.h
```diff
@@ -2,14 +2,26 @@
...
+/* 
+	Macro for registering an Aikido handler in the HOOKED_FUNCTIONS map.
+	It takes as parameters the PHP function name to be hooked and a C++ function 
+	that should be called BEFORE that PHP function is executed.
+	This macro doesn't register any hook for AFTER the function is execute, that's why
+	the second argument is nullptr.
+	The last nullptr part is a placeholder where the original function handler from
+	the Zend framework will be stored at initialization when we run the hooking.
+*/
+#define AIKIDO_REGISTER_FUNCTION_HANDLER_EX(function_name, pre_handler) AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(function_name, pre_handler, nullptr)
```
willem-delbare, line 24: macro is called post, but takes an argument called pre? is this correct?

lib/request-processor/context/context.go
```diff
@@ -3,37 +3,46 @@ package context
...
+	UserName                  *string
+	Body                      *string
+	BodyParsed                *map[string]string
+	Query                     *string
+	QueryParsed               *map[string]string
+	Cookies                   *string
+	CookiesParsed             *map[string]string
+	Headers                   *map[string]interface{}
+	HeadersParsed             *map[string]string
+	OutgoingRequestHostname   *string
```
willem-delbare: let's comment how we expect this state to behave. This context can chnage many times during the request, correct? (vs above is all static during entire reuqest). Let's make that clear somehow (at least in comments)

lib/request-processor/helpers/tryResolveToPrivateIp.go
```diff
@@ -0,0 +1,13 @@
+package helpers
+
+import "net"
+
+func TryResolveToPrivateIp(hostname string) string {
+	resolvedIps, _ := net.LookupHost(hostname)
```
willem-delbare: what's actually happening here below the hood? We do our own dns call to the os? let's comment that if so.

I assume this is a blocking operation. Can we add a (very short) timeout?

lib/request-processor/vulnerabilities/path-traversal/checkContextForPathTraversal.go
```diff
@@ -8,16 +8,18 @@ import (
...
+			inputString := SanitizePath(str)
+			if detectPathTraversal(sanitizedPath, inputString, checkPathStart) {
 				return &utils.InterceptorResult{
 					Operation:     operation,
 					Kind:          utils.Path_traversal,
 					Source:        source.Name,
 					PathToPayload: path,
 					Metadata: map[string]string{
-						"filename": filename,
+						"filename": sanitizedPath,
```
willem-delbare: this is for reporting right, so we might want the original input?

lib/request-processor/vulnerabilities/ssrf/findHostnameInUserInput.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	variants := []string{userInput, "http://" + userInput, "https://" + userInput}
+
+	for _, variant := range variants {
+		userInputURL := helpers.TryParseURL(variant)
+		if userInputURL != nil && userInputURL.Hostname() == hostnameURL.Hostname() {
+			userPort := helpers.GetPortFromURL(userInputURL)
+
+			if port == 0 {
```
willem-delbare, line 25: when would this happen? lets comment

lib/request-processor/vulnerabilities/ssrf/findHostnameInUserInput_test.go
```diff
@@ -0,0 +1,45 @@
...
+		{"http://", "localhost", 0, false},
+		{"localhost", "localhost localhost", 0, false},
+		{"http://169.254.169.254/latest/meta-data/", "169.254.169.254", 0, true},
+		{"http://2130706433", "2130706433", 0, true},
+		{"http://127.1", "127.1", 0, true},
+		{"http://127.0.1", "127.0.1", 0, true},
+		{"http://localhost", "localhost", 8080, false},
+		{"http://localhost:8080", "localhost", 8080, true},
+		{"http://localhost:8080", "localhost", 0, true},
+		{"http://localhost:8080", "localhost", 4321, false},
```
willem-delbare, line 35: there's no test with an actual domain being matchedi think

lib/request-processor/vulnerabilities/ssrf/imds.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func IsTrustedHostname(hostname string) bool {
+	// Check if the hostname is in the trusted hosts map
+	_, exists := trustedHosts[hostname]
+	return exists
+}
+
+func ResolvesToIMDSIp(hostname string, resolvedIps string) {
+
```
willem-delbare, line 36: not implemented?

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,74 @@
+package ssrf
+
+import (
+	"main/context"
+	"main/helpers"
+	"main/utils"
+)
+
+func CheckContextForSSRF(hostname string, port int, operation string) *utils.InterceptorResult {
```
willem-delbare, line 9: let's add when this is called. It's before or after the request?

lib/request-processor/vulnerabilities/ssrf/containsPrivateIPAddress_test.go
```diff
@@ -0,0 +1,202 @@
...
+	"fc00::",
+	"fc00::1",
+	"fc00::abc:1",
+	"fdff:ffff:ffff:ffff:ffff:ffff:ffff:ffff",
+	//"2130706433",
+	//"0x7f000001",
+
+	// AWS metadata
+	"fd00:ec2::254",
+	"169.254.169.254",
```
willem-delbare, line 160: probably you want localhost in this as well right?

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,74 @@
...
+				interceptorResult := utils.InterceptorResult{
+					Operation:     operation,
+					Kind:          utils.Ssrf,
+					Source:        source.Name,
+					PathToPayload: path,
+					Metadata:      getMetadataForSSRFAttack(hostname, port),
+					Payload:       str,
+				}
+
+				if containsPrivateIPAddress(hostname) {
```
willem-delbare, line 25: let's comment on the kind of ssrf we block here (probably the very obvious kind)

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,74 @@
...
+					return &interceptorResult
+				}
+
+				resolvedPrivateIp := helpers.TryResolveToPrivateIp(hostname)
```
willem-delbare: let's comment here what kind of case we detect here as well. It's when you create a cname that resolves to a private IP? or also 301 redirects?

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,74 @@
...
+				if resolvedPrivateIp != "" {
+					interceptorResult.Metadata["resolvedIp"] = resolvedPrivateIp
+					return &interceptorResult
+				}
+
+				context.ContextSetPartialInterceptorResult(interceptorResult)
```
willem-delbare: let's add comment what this does

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,74 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func CheckEffectiveHostnameForSSRF(effectiveHostname string) *utils.InterceptorResult {
```
willem-delbare, line 56: comment how the effecitve hostname is different. This is for curl only or also for others? It's the url after redirects?

lib/php-extension/include/Utils.h
tudor-timcu: Yes, it's ok. The AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX takes both PRE handler and POST handler. The AIKIDO_REGISTER_FUNCTION_HANDLER_EX registers only the preHandler. I reuse the macros for easier maintenance

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil
+	}
+
+	// After redirects the effective hostname resolved to a private IP -> SSRF
+	interceptorResult.Metadata["effectiveHostname"] = effectiveHostname
+	interceptorResult.Metadata["resolvedIp"] = resolvedPrivateIp
+	return interceptorResult
+}
+
+func CheckResolvedIpForSSRF(resolvedIp string) *utils.InterceptorResult {
```
willem-delbare, line 81: let's comment what the coverage is for this one. only curl? or only file_get_contents?

lib/php-extension/Handle.cpp
```diff
@@ -19,30 +19,30 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
...
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(touch, handle_file_path_access),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(unlink, handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chdir, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chgrp, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chmod, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chown, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(copy, handle_pre_file_path_access_2, handle_post_file_path_access_2),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file, handle_pre_file_path_access, handle_post_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file_get_contents, handle_pre_file_path_access, handle_post_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(file_put_contents, handle_pre_file_path_access, handle_post_file_path_access),
```
willem-delbare: the post handler is for ssrf, but i don't think this function needs ssrf protection

lib/request-processor/context/cache.go
willem-delbare: comments doesn't explain what it is. Can you give example of the data insidE?

lib/request-processor/handle_urls.go
```diff
@@ -1,33 +1,55 @@
...
 import (
+	"main/attack"
 	"main/context"
 	"main/grpc"
 	"main/log"
 	"main/utils"
+	ssrf "main/vulnerabilities/ssrf"
 )
 
 func OnPreFunctionExecutedCurl() string {
```
willem-delbare: let's comment what the goal here is: defends vs normal , direct SSRF attacks, with a direct IP address inserted or also other stuff ?

lib/request-processor/handle_urls.go
```diff
@@ -1,33 +1,55 @@
...
+	log.Info("[BEFORE] Got domain: ", hostname)
 	//TODO: check if domain is blacklisted
 	return ""
 }
 
-func OnAfterFunctionExecutedCurl() string {
-	url := context.GetOutgoingRequestUrl()
-	port := context.GetOutgoingRequestPort()
-	if url == "" {
+func OnPostFunctionExecutedCurl() string {
```
willem-delbare: let's comment the goal here: i assume the goal is to downgrade a SSRF, in scase it already took place, to a blind SSRF by not returning the data and throwing an exception. If so let's comment that

lib/request-processor/handle_urls.go
```diff
@@ -1,33 +1,55 @@
...
-	go grpc.OnDomain(domain, port)
+	log.Info("[AFTER] Got domain: ", hostname, " port: ", port)
+
+	var res *utils.InterceptorResult = nil
+	go grpc.OnDomain(hostname, port)
+	if effectiveHostname != hostname {
+		// After the request was made, if effective hostname is different that the initially requested one (redirects, ...)
+		// -> check for SSRF
+		// -> report it to the agent
+		res = ssrf.CheckEffectiveHostnameForSSRF(effectiveHostname)
```
willem-delbare: if we can check the resolved ip , why do we need to check the effective hostname?

or do some functions support one and not the other? If so, let's comment with examples

lib/request-processor/context/cache.go
willem-delbare: Ok can we rename it to httpcall interceptor result?
So this wont accidentally be used for other things?

lib/php-extension/Handle.cpp
```diff
@@ -19,30 +19,30 @@ unordered_map<std::string, PHP_HANDLERS> HOOKED_FUNCTIONS = {
...
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(rmdir, handle_file_path_access),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(scandir, handle_file_path_access),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(symlink, handle_file_path_access_2),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(touch, handle_file_path_access),
-	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(unlink, handle_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chdir, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chgrp, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chmod, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_EX(chown, handle_pre_file_path_access),
+	AIKIDO_REGISTER_FUNCTION_HANDLER_WITH_POST_EX(copy, handle_pre_file_path_access_2, handle_post_file_path_access_2),
```
willem-delbare: this usage probably too niche, don't think we need to support it

willem-delbare: (for ssrf)

lib/request-processor/context/context.go
```diff
@@ -124,24 +136,41 @@ func GetUserName() string {
 	return GetFromCache(ContextSetUserName, &Context.UserName)
 }
 
-func GetFunctionName() string {
-	return Context.Callback(C.FUNCTION_NAME)
+func GetHostNameAndPort(urlCallbackOption int) (string, int) {
```
willem-delbare: what is a callback option? I don't really get what the goal of this function is or when/why it's used

lib/request-processor/context/context.go
```diff
@@ -124,24 +136,41 @@ func GetUserName() string {
...
-		return 0
-	}
-	return port
+func GetOutgoingRequestEffectiveHostnameAndPort() (string, int) {
+	return GetHostNameAndPort(C.OUTGOING_REQUEST_EFFECTIVE_URL)
+}
+
+func GetOutgoingRequestResolvedIp() string {
+	return Context.Callback(C.OUTGOING_REQUEST_RESOLVED_IP)
+}
```
willem-delbare: what is this callback function? Why don't we just return the data?

lib/request-processor/context/context.go
```diff
@@ -29,11 +32,20 @@ type ContextData struct {
 	CookiesParsed *map[string]string
 	Headers       *map[string]interface{}
 	HeadersParsed *map[string]string
+
+	/* Event level context cache below (changes on each PHP function call) */
+	OutgoingRequestHostname   *string
+	OutgoingRequestPort       *int
+	OutgoingRequestResolvedIp *string
+	PartialInterceptorResult  *utils.InterceptorResult
```
willem-delbare: currentSSRFInterceptorResults ?

lib/request-processor/handle_urls.go
```diff
@@ -1,33 +1,77 @@
...
+		if res == nil {
+			// We double check here for SSRF on the effective hostname because some sinks might not provide the resolved IP address
+			res = ssrf.CheckEffectiveHostnameForSSRF(effectiveHostname)
+		}
+		go grpc.OnDomain(effectiveHostname, effectivePort)
+	}
+
+	if res != nil {
+		go grpc.OnAttackDetected(*res)
+		return attack.GetAttackDetectedAction(*res)
```
willem-delbare, line 82: ah, what do we actually do when the attack is detected? We throw? or we return '' ?

willem-delbare: if we throw before we return, let's comment here that that is the goal.

I guess we're actually not downgrading to blind ssrf, we're actually stopping it

lib/request-processor/helpers/resolveHostname.go
```diff
@@ -0,0 +1,28 @@
...
+	"net"
+	"time"
+)
+
+/*
+This function tries to resolve the hostname to a private IP adress, if possible.
+It does this by calling DNS resolution from the OS (getaddrinfo for Linux).
+*/
+func ResolveHostname(hostname string) []string {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Millisecond)
```
willem-delbare: let's comment what happens if the timeout is reached?

also 5 ms is maybe a bit short. 50?

lib/request-processor/vulnerabilities/ssrf/getResolvedIpStatus.go
```diff
@@ -0,0 +1,23 @@
...
+
+type ResolvedIpStatus struct {
+	ip        string
+	isPrivate bool
+	isIMDS    bool
+}
+
+func getResolvedIpStatusForHostname(hostname string) *ResolvedIpStatus {
+	resolvedIps := helpers.ResolveHostname(hostname)
+	imdsIP := TryGetIMDSIp(hostname, resolvedIps)
```
willem-delbare: tryGet is an odd function name. What is the intent of this function?

lib/request-processor/vulnerabilities/ssrf/getResolvedIpStatus.go
```diff
@@ -0,0 +1,23 @@
...
+	if imdsIP != "" {
+		return &ResolvedIpStatus{ip: imdsIP, isPrivate: false, isIMDS: true}
+	}
+
+	privateIp := helpers.TryGetPrivateIp(resolvedIps)
```
willem-delbare: also not a great funciton name, Could either rename of comment what the intent is

lib/request-processor/vulnerabilities/ssrf/imds.go
```diff
@@ -0,0 +1,43 @@
...
+	return false
+}
+
+func isTrustedHostname(hostname string) bool {
+	// Check if the hostname is in the trusted hosts map
+	_, exists := trustedHosts[hostname]
+	return exists
+}
+
+func TryGetIMDSIp(hostname string, resolvedIps []string) string {
```
willem-delbare: maybe something like 'isHostnameAndIPIMDSAttack'

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,96 @@
...
+					Payload:       str,
+				}
+
+				if containsPrivateIPAddress(hostname) {
+					// Hostname was found in user input and is actually a private IP address (http://127.0.0.1) -> SSRF
+					interceptorResult.Metadata["isPrivateIp"] = "true"
+					return &interceptorResult
+				}
+
+				resolvedIpStatus := getResolvedIpStatusForHostname(hostname)
```
willem-delbare, line 31: let's not that we on purpose introduce our own dns request here, and that while we expect delays, we assume this is cached (let's unit test this?)

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,96 @@
...
+				if resolvedIpStatus != nil {
+					interceptorResult.Metadata["resolvedIp"] = resolvedIpStatus.ip
+					if resolvedIpStatus.isIMDS {
+						// Hostname was found in user input and it resolves to an IMDS IP address -> stored SSRF
```
willem-delbare: for the stored SSRF protection this needs to be outside of the findHostnameInUserInput() block, so this comment is not correct

willem-delbare: (but code is ok)

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,96 @@
...
+/* This is called after the request is made to check for SSRF in the effective hostname - hostname optained after redirects from the PHP library that made the request (curl) */
+func CheckEffectiveHostnameForSSRF(effectiveHostname string) *utils.InterceptorResult {
+	interceptorResult := context.GetPartialInterceptorResult()
+	if interceptorResult == nil {
+		// The initially requested hostname was not found in the user input -> no SSRF
+		return nil
+	}
+
+	interceptorResult.Metadata["effectiveHostname"] = effectiveHostname
+	resolvedIpStatus := getResolvedIpStatusForHostname(effectiveHostname)
```
willem-delbare, line 64: ok let's comment here too that we expect this to be cached (and unit test)

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,96 @@
...
+	if resolvedIpStatus != nil {
+		interceptorResult.Metadata["resolvedIp"] = resolvedIpStatus.ip
+		if resolvedIpStatus.isIMDS {
+			// Hostname was found in user input and the effective hostname (after redirects) resolved to an IMDS IP address -> stored SSRF
```
willem-delbare: comment is copy pastede, not correct

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+/* This is called after the request is made to check for SSRF in the effective hostname - hostname optained after redirects from the PHP library that made the request (curl) */
+func CheckEffectiveHostnameForSSRF(effectiveHostname string) *utils.InterceptorResult {
+	interceptorResult := context.GetPartialInterceptorResult()
+	if interceptorResult == nil {
+		// The initially requested hostname was not found in the user input -> no SSRF
+		return nil
+	}
```
willem-delbare, line 61: this is fairly critical stuff. Let's do an e2e test that checks both the AWS SDK and the GCP sdk can still do their respective fetches without being blocked

lib/request-processor/vulnerabilities/ssrf/checkContextForSSRF.go
tudor-timcu: There are 2 E2E tests checking this.

lib/request-processor/handle_urls.go
```diff
@@ -1,33 +1,85 @@
...
+	go grpc.OnDomain(hostname, port)
+
+	res := ssrf.CheckResolvedIpForSSRF(resolvedIp)
+	if effectiveHostname != hostname {
+		// After the request was made, the effective hostname is different that the initially requested one (redirects)
+		if res == nil {
+			// We double check here for SSRF on the effective hostname because some sinks might not provide the resolved IP address
+			res = ssrf.CheckEffectiveHostnameForSSRF(effectiveHostname)
+		}
+		go grpc.OnDomain(effectiveHostname, effectivePort)
```
willem-delbare, line 75: i don't think we need to report those redirect domains as well for that list

lib/request-processor/context/context.go
willem-delbare: lett's comment why that is

lib/request-processor/context/context.go
willem-delbare: (lets comment why that is)

lib/request-processor/context/context.go
```diff
@@ -3,13 +3,16 @@ package context
...
-	"strconv"
+	"main/utils"
+	"net/url"
 )
 
 type CallbackFunction func(int) string
 
 type ContextData struct {
+	/* Request level context cache below (changes on each PHP request) */
 	Callback      CallbackFunction
```
willem-delbare: let's comment what is in here? What is a callbackfunction? what is used for?

lib/request-processor/context/event_getters.go
```diff
@@ -0,0 +1,65 @@
...
+
+func GetSqlDialect() string {
+	return Context.Callback(C.SQL_DIALECT)
+}
+
+func GetModule() string {
+	return Context.Callback(C.MODULE)
+}
+
+func getHostNameAndPort(urlCallbackId int) (string, int) {
```
willem-delbare:
```suggestion
func getHostNameAndPort(urlCallbackId int) (string, int) { // urlcallbackid is the type of data we request, eg C.OUTGOING_REQUEST_URL
```

PR #57, opened 2024-10-01
API discovery

tudor-timcu:
- Added API discovery support in Request Processor
- Send API schema along with request metadata via gRPC to Agent
- Added API schema merge support in Agent
- Added unitests for api discovery
- Added E2E tests for api discovery

### DIFF
--- a/lib/agent/cloud/event_heartbeat.go
+++ b/lib/agent/cloud/event_heartbeat.go
@@ -37,6 +37,8 @@ func GetRoutesAndClear() []Route {
 		}
 	}
 
+	// Clear routes data
+	globals.Routes = make(map[string]map[string]*Route)
 	return routes
 }
 
--- a/lib/request-processor/context/context_for_unit_tests.go
+++ b/lib/request-processor/context/context_for_unit_tests.go
@@ -45,7 +45,8 @@ func LoadForUnitTests(context map[string]string) {
 }
 
 func UnloadForUnitTests() {
-	Context = ContextData{}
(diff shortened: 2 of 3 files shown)


lib/request-processor/context/data_sources.go
```diff
@@ -6,8 +6,8 @@ type Source struct {
 }
 
 var SOURCES = []Source{
-	{"body", GetParsedBody},
-	{"query", GetParsedQuery},
-	{"headers", GetParsedHeaders},
-	{"cookies", GetParsedCookies},
+	{"body", GetBodyParsedWithPathToPayload},
```
willem-delbare: i assume this happens at the start of the request. How do we already know the attack payload?

lib/agent/cloud/event_heartbeat.go
```diff
@@ -27,13 +27,16 @@ func GetRoutesAndClear() []Route {
...
+		for _, routeData := range methodsMap {
+			if routeData.Hits == 0 {
+				continue
+			}
+			routes = append(routes, *routeData)
+			routeData.Hits = 0
 		}
 	}
 
-	globals.Routes = make(map[string]map[string]int)
```
willem-delbare, line 36: isn't this a reset after the heartbeat? why would we no longer need it?

willem-delbare: let's follow the node way (i guess it does resets?) and let's add a comment that we are explicitly doing that reset with this line

PR #63, opened 2024-10-09
nginx (php-fpm) & apache (mod-php) automated E2E testing

tudor-timcu:
- nginx (php-fpm) automated E2E testing with the existing E2E tests (25 tests)
- apache (mod-php) automated E2E testing with the existing E2E tests (25 tests)
- fixed all related issues found in the E2E testing with both these servers
- fixed exit handler for PHP 8 (we now use the same exit mechanism as PHP 7)
- improved header extraction from PHP layer to also get headers that don't start with HTTP_ in all cases (like Content-Type, Authorization, Content-Length, ...)


lib/php-extension/Utils.cpp
```diff
@@ -188,13 +207,13 @@ bool aikido_echo(std::string message) {
 }
 
 bool aikido_exit() {
-#if PHP_VERSION_ID >= 80000
-    int _result = zend_eval_stringl("exit();", strlen("exit();"), NULL, "aikido php code (exit action)");
-    AIKIDO_LOG_INFO("Called 'exit' eval -> result %d\n", _result == SUCCESS);
-    return _result == SUCCESS;
-#else
+//#if PHP_VERSION_ID >= 80000
+//    int _result = zend_eval_stringl("exit();", strlen("exit();"), NULL, "aikido php code (exit action)");
```
willem-delbare: Is this on purpose?

PR #88, opened 2024-11-19
Store monitored sink stats

tudor-timcu:


lib/php-extension/Environment.cpp
```diff
@@ -62,4 +78,5 @@ void LoadEnvironment() {
     AIKIDO_GLOBAL(token) = GetEnvString("AIKIDO_TOKEN", "");
     AIKIDO_GLOBAL(endpoint) = GetEnvString("AIKIDO_ENDPOINT", "https://guard.aikido.dev/");
     AIKIDO_GLOBAL(config_endpoint) = GetEnvString("AIKIDO_REALTIME_ENDPOINT", "https://runtime.aikido.dev/");
+    AIKIDO_GLOBAL(report_stats_interval) = GetEnvNumber("AIKIDO_REPORT_STATS_INTERVAL", 10000);
```
willem-delbare: too rapid

lib/php-extension/Environment.cpp
```diff
@@ -1,5 +1,7 @@
 #include "Includes.h"
 
+#define MIN_REPORT_STATS_INTERVAL 50
```
willem-delbare: hmm report to where? is this seconds or minutes?

lib/php-extension/include/php_aikido.h
```diff
@@ -17,6 +17,7 @@ bool disable;
 bool collect_api_schema;
 bool trust_proxy;
 bool localhost_allowed_by_default;
+unsigned int report_stats_interval;
```
willem-delbare: let's add comments? Report to agent or to cloud?

PR #93, opened 2024-11-25
Handle shell execution with hashtable as first parameter

tudor-timcu:


lib/php-extension/HandleShellExecution.cpp
```diff
@@ -12,7 +31,36 @@ AIKIDO_HANDLER_FUNCTION(handle_shell_execution) {
...
+    else if (Z_TYPE_P(cmdVal) == IS_ARRAY) {
+        HashTable* cmdTokens = Z_ARRVAL_P(cmdVal);
+        if (!cmdTokens) {
+            return;
+        }
+        std::string cmdString = get_command(cmdTokens);
+        if (cmdString.empty()) {
+            return;
+        }
+        helper_handle_pre_shell_execution(cmdString, eventId);
```
willem-delbare: Already checked on line 51 right?

PR #97, opened 2024-12-04
Laravel environment support

tudor-timcu:


lib/request-processor/main.go
```diff
@@ -86,6 +86,21 @@ func RequestProcessorContextInit(contextCallback C.ContextCallback) (initOk bool
 	return context.Init(GoContextCallback)
 }
 
+//export RequestProcessorConfigUpdate
```
willem-delbare, line 92: let's add comments (goal of this function, when to use)

PR #109, opened 2024-12-17
PHP filters support

tudor-timcu:


lib/php-extension/HandlePathAccess.cpp
```diff
@@ -2,8 +2,10 @@
 
 /* Helper for handle pre file path access */
 void helper_handle_pre_file_path_access(char *filename, EVENT_ID &eventId) {
-    if (strncmp(filename, "php://", 6) == 0) {
-        // Whitelist php:// streams as they are often used by PHP frameworks a lot
+    if (strncmp(filename, "php://", 6) == 0 &&
```
kapyteinaikido, line 5: This could still be dangerous if the developer enabled the following PHP.ini config settings: `allow_url_fopen` and  `allow_url_include`.

A vulnerable instance would look like the following:

```
<?php

include($_GET['page']);

?>
```

Malicious payload would be:

```
http://localhost:1337/?page=php://input
```
with HTTP body:
```
<?php echo 1; ?>
```

tudor-timcu: Discussed that we won't handle this atm because the user needs to enable a lot of things to make this bypass happen. We'll just report the state of these settings back to dashboard and notify the user to not enable these.

docs/should_block_request.md
```diff
@@ -30,12 +30,13 @@ class AikidoMiddleware implements MiddlewareInterface
...
-        // Get the user ID (from session or other auth system)
+        // Get the user ID / name (from session or other auth system)
         $userId = $this->getAuthenticatedUserId();
+        $userName = $this->getAuthenticatedUserName();
         
         // If the user is authenticated, set the user ID in Aikido Zen context
-        if ($userId) {
-            \aikido\set_user($userId);
+        if ($userId && $userName) {
+            \aikido\set_user(strval($userId), $userName);
```
willem-delbare:
```suggestion
            \aikido\set_user( $userId, $userName);
```

PR #115, opened 2024-12-20
Wildcard config support for routes / methods with support for rate limiting the most restrictive match

tudor-timcu:
- Added support for configs that use wildcards when adding routes / method in the Zen dashboard
- Updated rate limiting logic to also maintain counts for the wildcard endpoints (if they match) and provide the rate limiting status back to the request process if threshold is exceeded
- Added support for selecting the most restrictive config from multiple matching ones:
    - If at least one force protection off is set to true in all the matching endpoints configs -> the protection is of
    - The allowedIPAddresses fields are checked for all the matching routes and if it's found in at least one -> the rest of the checks are bypassed
    - Same for checking if the rate limiting is enabled for the current route, if at least one config has rate limiting enabled for the matching endpoints -> we do the gRPC request to get the rate limiting status from agent
- Added E2E tests to check these config selection cases + rate limiting on wildcard routes


lib/agent/cloud/common.go
```diff
@@ -76,20 +83,33 @@ func UpdateRateLimitingConfig() {
...
+		rateLimitingValue := &RateLimitingValue{
 			Config: RateLimitingConfig{
 				MaxRequests:         newEndpointConfig.RateLimiting.MaxRequests,
 				WindowSizeInMinutes: newEndpointConfig.RateLimiting.WindowSizeInMS / MinRateLimitingIntervalInMs},
 			UserCounts: make(map[string]*RateLimitingCounts),
 			IpCounts:   make(map[string]*RateLimitingCounts),
 		}
+
+		if isWildcardEndpoint(k.Route) {
+			routeRegex, err := regexp.Compile(strings.ReplaceAll(k.Route, "*", ".*"))
```
bitterpanda63: How does it handle trailing slashes (In node we append this at the end: `/?`, i.e. optional trailing slash)

bitterpanda63: For uniformity, I know it's just a capture group, but makes it easier to debug if route string is not all good
```suggestion
			routeRegex, err := regexp.Compile(strings.ReplaceAll(k.Route, "*", "(.*)"))
```

lib/agent/grpc/request.go
```diff
@@ -148,26 +145,53 @@ func isRateLimitingThresholdExceeded(config *RateLimitingConfig, countsMap map[s
...
-			log.Infof("Rate limited request for ip %s - %s %s - %v", ip, method, route, rateLimitingDataForRoute.IpCounts[ip])
-			return &protos.RateLimitingStatus{Block: true, Trigger: "ip"}
+	}
+	return wildcardRatelimitingValues
+}
+
+func getMatchingRateLimitingValues(method, route string) []*RateLimitingValue {
+	rateLimitingDataArray := getRateLimitingValue(method, route)
+	rateLimitingDataArray = append(rateLimitingDataArray, getRateLimitingValue("*", route)...)
+	rateLimitingDataArray = append(rateLimitingDataArray, getWildcardRateLimitingValues(method, route)...)
```
bitterpanda63, line 196: You prioritize the defined method over "*" for non-wildcard routes but I don't see a sort for wildcard routes, i.e. even on wildcards a well-defined method is prioritized over a wildcard method

bitterpanda63: In node : 
```ts
  possible.sort((a, b) => {
    if (a.method === b.method) {
      return 0;
    }

    if (a.method === "*") {
      return 1;
    }

    return -1;
  });
```

tests/server/test_user_rate_limiting_1_minute_for_wildcard_route/env.json
```diff
@@ -0,0 +1,5 @@
+{
+    "AIKIDO_BLOCK": "1",
+    "AIKIDO_LOCALHOST_ALLOWED_BY_DEFAULT": "0",
+    "AIKIDO_FEATURE_COLLECT_API_SCHEMA": "1"
+}
```
bitterpanda63, line 5: Please check .editorconfig, always end with newline

lib/request-processor/utils/config.go
```diff
@@ -5,6 +5,30 @@ import (
...
+		return []WildcardEndpointData{}
+	}
+	return wildcardRoutesForMethod
+}
+
+func GetWildcardEndpointsConfigs(method string, route string) []EndpointData {
+	globals.CloudConfigMutex.Lock()
+	defer globals.CloudConfigMutex.Unlock()
+
+	wildcardRoutes := GetWildcardEndpointsConfigsForMethod(method)
```
bitterpanda63, line 21:
```suggestion
    // We prioritize defined methods over wildcard methods
	wildcardRoutes := GetWildcardEndpointsConfigsForMethod(method)
```

tests/server/test_user_rate_limiting_1_minute_for_wildcard_route/env.json
hansott: Yes, would be nice to add one! 🥳

tudor-timcu: I'll do this in a separate PR.

lib/agent/grpc/request.go
```diff
@@ -148,25 +149,78 @@ func isRateLimitingThresholdExceeded(config *RateLimitingConfig, countsMap map[s
...
+
+func getRateLimitingStatus(method, route, user, ip string) *protos.RateLimitingStatus {
 	globals.RateLimitingMutex.RLock()
 	defer globals.RateLimitingMutex.RUnlock()
 
-	rateLimitingDataForRoute, exists := globals.RateLimitingMap[RateLimitingKey{Method: method, Route: route}]
-	if !exists {
-		return &protos.RateLimitingStatus{Block: false}
-	}
+	rateLimitingDataMatch := getRateLimitingDataForEndpoint(method, route)
```
hansott: Check if `nil`?

lib/request-processor/utils/utils.go
```diff
@@ -127,27 +127,26 @@ func isLocalhost(ip string) bool {
...
+		return -1
 	}
-
 	if allowedIps.Contains(ipAddress) {
-		return true
+		return 1
 	}
 
-	return false
+	return 0
```
hansott: why did we change to int? a bit unclear what each int means, can we simplify this? perhaps an enum?

lib/agent/cloud/common.go
```diff
@@ -42,6 +44,9 @@ func ResetHeartbeatTicker() {
 		}
 	}
 }
+func isWildcardEndpoint(route string) bool {
+	return strings.Contains(route, "*")
+}
```
hansott, line 50: use utils fn?

lib/agent/grpc/request.go
```diff
@@ -148,25 +149,78 @@ func isRateLimitingThresholdExceeded(config *RateLimitingConfig, countsMap map[s
...
+	wildcardMethodRateLimitingData := getRateLimitingValue("*", route)
+	if wildcardMethodRateLimitingData != nil {
+		rateLimitingDataArray = append(rateLimitingDataArray, wildcardMethodRateLimitingData)
+	}
+	rateLimitingDataArray = append(rateLimitingDataArray, getWildcardRateLimitingValues(method, route)...)
+	rateLimitingDataArray = append(rateLimitingDataArray, getWildcardRateLimitingValues("*", route)...)
+	return rateLimitingDataArray
+}
+
+func getRateLimitingDataForEndpoint(method, route string) *RateLimitingValue {
```
hansott: We should match the wildcards against the path of the `url` instead of route: https://github.com/AikidoSec/firewall-node/blob/main/library/helpers/matchEndpoints.ts#L43

hansott: We also sort the endpoints based on the amount of `*`, so that we first match endpoints with the least wildcard routes: https://github.com/AikidoSec/firewall-node/blob/main/library/helpers/matchEndpoints.ts#L46-L49

lib/request-processor/handle_blocking_request.go
```diff
@@ -68,7 +67,12 @@ func OnGetBlockingStatus() string {
 		return GetStoreAction("blocked", "user-agent", userAgentBlockedDescription, userAgent)
 	}
 
-	if endpointData != nil && endpointData.RateLimiting.Enabled {
+	if userAgentBlocked, userAgentBlockedDescription := utils.IsUserAgentBlocked(userAgent); userAgentBlocked {
+		log.Infof("User Agent \"%s\" blocked due to: %s!", userAgent, userAgentBlockedDescription)
+		return GetStoreAction("blocked", "user-agent", userAgentBlockedDescription, userAgent)
+	}
```
hansott: Seems like we have some duplicate code here? See above

lib/agent/grpc/request.go
tudor-timcu: route already is the url path at that point: see https://github.com/AikidoSec/firewall-php/blob/19f9f824621f5a31607b2f9c2602b88640a98044/lib/request-processor/utils/build_route_from_url.go#L22

tudor-timcu: Fixed the sorting

hansott (approved):

PR #120, opened 2025-01-03
Align shouldDiscoverRoute to Node version

timokoessler:
Discover redirects, ignore all errors

### DIFF
--- a/lib/request-processor/utils/should_discover_route.go
+++ b/lib/request-processor/utils/should_discover_route.go
@@ -15,14 +15,6 @@ const (
 	PERMANENT_REDIRECT = 308
 )
 
-var ERROR_CODES = []int{NOT_FOUND, METHOD_NOT_ALLOWED}
-var REDIRECT_CODES = []int{
-	MOVED_PERMANENTLY,
-	FOUND,
-	SEE_OTHER,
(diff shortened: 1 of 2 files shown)


hansott (approved):

lib/request-processor/utils/should_discover_route.go
```diff
@@ -33,11 +25,7 @@ func ShouldDiscoverRoute(statusCode int, route, method string) bool {
 		return false
 	}
 
-	if containsInt(ERROR_CODES, statusCode) {
-		return false
-	}
-
-	if containsInt(REDIRECT_CODES, statusCode) {
+	if statusCode < 200 || statusCode > 399 {
```
hansott, line 28: Seems like we can combine the segments below? or maybe another PR? We loop over the segments multiple times:

```go
for _, segment := range segments {
		if isDotFile(segment) {
			return false
		}
	}

	for _, segment := range segments {
		if containsIgnoredString(segment) {
			return false
		}
	}

	for _, segment := range segments {
		if !isAllowedExtension(segment) {
			return false
		}
	}
```

timokoessler: Changed.

tudor-timcu (approved):

PR #141, opened 2025-01-27
Added support for PDO statement execute

tudor-timcu:
- Added support for PDO statement execute
- Added support for URL decode of query paramenters & form-data body
- Added support for URL query array parameters


lib/request-processor/utils/utils.go
```diff
@@ -52,6 +53,10 @@ func ParseFormData(data string, separator string) map[string]interface{} {
 			continue
 		}
 		result[keyValue[0]] = keyValue[1]
+		decodedValue, err := url.QueryUnescape(keyValue[1])
```
kapyteinaikido, line 56: Would have to check if PHP's urldecode(...) is as forgiving as `QueryUnescape`.

lib/request-processor/utils/utils.go
```diff
@@ -72,6 +77,11 @@ func ParseBody(body string) map[string]interface{} {
 }
 
 func ParseQuery(query string) map[string]interface{} {
+	jsonQuery := map[string]interface{}{}
```
willem-delbare, line 80: can we add comments about what is going on here?

lib/php-extension/HandlePDO.cpp
```diff
@@ -35,3 +35,28 @@ AIKIDO_HANDLER_FUNCTION(handle_pre_pdo_query) {
...
+    }
+
+    pdo_stmt_t *stmt = Z_PDO_STMT_P(pdo_statement_object);
+    if (!stmt->dbh) { // object is not initialized 
+        return;
+    }
+
+    eventId = EVENT_PRE_SQL_QUERY_EXECUTED;
+    eventCache.moduleName = "PDOStatement"; 
+    eventCache.sqlDialect = "unknown";
```
willem-delbare, line 52: ?

lib/php-extension/Request.cpp
```diff
@@ -76,6 +76,48 @@ std::string Request::GetBody() {
     return "";
 }
 
+/**
+ * Converts the current HTTP query parameters (_GET) into a JSON-formatted string.
+ *
+ * This function retrieves the query parameters from the `_GET` global array in PHP 
+ * and constructs a JSON object representation of the parameters. It supports both 
+ * scalar values (e.g., "key=value") and array values (e.g., "key[]=value1&key[]=value2").
```
willem-delbare, line 84: what's the reason we do this here?

lib/request-processor/utils/utils.go
```diff
@@ -72,6 +77,11 @@ func ParseBody(body string) map[string]interface{} {
...
+	err := json.Unmarshal([]byte(query), &jsonQuery)
```
kapyteinaikido, line 81: This will fail if invalid Unicode is found in the query parameter (which is fine for PHP), meaning we'll lose all context.

lib/php-extension/Request.cpp
```diff
@@ -76,6 +76,48 @@ std::string Request::GetBody() {
...
+*/
+std::string Request::GetQuery() {
+  
+    zval *get_array;
+    get_array = zend_hash_str_find(&EG(symbol_table), "_GET", sizeof("_GET") - 1);
```
willem-delbare, line 92: -

PR #150, opened 2025-02-17
Token and env improvements

tudor-timcu:


lib/agent/grpc/server.go
```diff
@@ -25,7 +25,7 @@ type server struct {
 
 func (s *server) OnConfig(ctx context.Context, req *protos.Config) (*emptypb.Empty, error) {
 	previousToken := config.GetToken()
-	if previousToken == "" {
+	if previousToken == "" && req.GetToken() != "" {
```
willem-delbare, line 28: let's add some comments. What is req. gettoken?

lib/php-extension/Environment.cpp
```diff
@@ -41,7 +41,7 @@ std::string GetEnvString(const std::string& env_key, const std::string default_v
 }
 
 bool GetEnvBool(const std::string& env_key, bool default_value) {
-    std::string env_value = GetEnvVariable(env_key);
+    std::string env_value = ToLowercase(GetEnvVariable(env_key));
```
willem-delbare, line 44: the key is not case sensitive?

lib/agent/cloud/common.go
```diff
@@ -140,3 +141,16 @@ func StoreCloudConfig(configReponse []byte) bool {
 	ApplyCloudConfig()
 	return true
 }
+
+func LogCloudRequestError(text string, err error) {
+	if atomic.LoadUint32(&globals.GotTraffic) == 0 {
+		return
+	}
+	if err.Error() == "no token set" {
```
willem-delbare: let's comment what the goal is. Feels like it swallows the error under certain conditions?

PR #171, opened 2025-03-18
Aikido Agent in a separate process

tudor-timcu:
- Start Aikido Agent in a separate process on init in order to fix Apache graceful restarts and not duplicate memory via subsequent forks
- On uninit, kill that process with SIGTERM (let the process gracefully stop)
- It passes the json configuration via command line args and the token via env variables


lib/agent/main.go
```diff
@@ -22,28 +24,47 @@ func AgentInit(initJson string) (initOk bool) {
...
 
-	log.Infof("Aikido Agent v%s unloaded!", globals.Version)
+	log.Infof("Aikido Agent v%s stopped!", globals.Version)
 	log.Uninit()
 }
 
-func main() {}
+func main() {
+	if len(os.Args) != 2 {
+		log.Errorf("Usage: %s <init_json>", os.Args[0])
```
willem-delbare, line 57: ?

willem-delbare (approved):

PR #176, opened 2025-03-21
Added support for multiple php versions installed on the same machine

tudor-timcu:


lib/request-processor/grpc/client.go
```diff
@@ -129,7 +129,7 @@ func GetCloudConfig() {
 
 	cloudConfig, err := client.GetCloudConfig(ctx, &protos.CloudConfigUpdatedAt{ConfigUpdatedAt: utils.GetCloudConfigUpdatedAt()})
 	if err != nil {
-		log.Infof("Could not get cloud config: %v", err)
+		//log.Infof("Could not get cloud config: %v", err)
```
willem-delbare: ?

PR #188, opened 2025-04-11
Fixed wordpress crash

tudor-timcu:


lib/php-extension/include/Server.h
```diff
@@ -1,17 +1,13 @@
 #pragma once
 
-class Request {
+class Server {
    private:
-    zval* server = nullptr;
+    zend_string* serverString = nullptr;
```
willem-delbare, line 7: let's comment what is stored in here

willem-delbare (approved):

PR #190, opened 2025-05-12
Auto inject middleware into AST (abstract syntax tree) before the PHP files are compiled

tudor-timcu:


lib/php-extension/Action.cpp
```diff
@@ -18,7 +18,8 @@ ACTION_STATUS Action::executeExit(json &event) {
     CallPhpFunctionWithOneParam("http_response_code", _response_code);
     CallPhpFunctionWithOneParam("header", "Content-Type: text/plain");
     CallPhpEcho(_message);
-    exit = true;
+
+    zend_bailout();
```
hansott, line 23: Would be nice to add a comment here? Why `zend_bailout`?

lib/php-extension/Aikido.cpp
```diff
@@ -48,6 +48,7 @@ PHP_MSHUTDOWN_FUNCTION(aikido) {
         return SUCCESS;
     }
 
+    UnhookZendAstProcess();
```
hansott: Needs to be above `AIKIDO_GLOBAL(sapi_name) == "cli")`? Since we hook before the CLI check?

lib/php-extension/Aikido.cpp
```diff
@@ -91,8 +93,10 @@ PHP_MINFO_FUNCTION(aikido) {
 
 static const zend_function_entry ext_functions[] = {
     ZEND_NS_FE("aikido", set_user, arginfo_aikido_set_user)
-        ZEND_NS_FE("aikido", should_block_request, arginfo_aikido_should_block_request)
-            ZEND_FE_END};
+    ZEND_NS_FE("aikido", should_block_request, arginfo_aikido_should_block_request)
+    ZEND_NS_FE("aikido", auto_block_request, arginfo_aikido_auto_block_request)
```
hansott, line 101: Should we use a name like `\aikido\__do_not_use_auto_block_request`? e.g. if people find the function using `get_defined_functions`

lib/php-extension/HandleShouldBlockRequest.cpp
```diff
@@ -1,12 +1,19 @@
 #include "Includes.h"
 
 zend_class_entry *blockingStatusClass = nullptr;
+bool checkedAutoBlock = false;
+bool checkedShouldBlockRequest = false;
```
hansott, line 11: Do we want to keep a separate state?

The auto block will be called before should block? So any existing calls to should block are not needed anymore?

lib/php-extension/HookAst.cpp
```diff
@@ -0,0 +1,123 @@
+#include "Includes.h"
+
+HashTable *global_ast_to_clean;
+ZEND_API void (*original_ast_process)(zend_ast *ast) = nullptr;
+
+void ast_to_clean_dtor(zval *zv){
```
hansott: Maybe add comment that this is called to clean up value of hash table when key is cleaned up? (if I understand it well enough)

lib/php-extension/HookAst.cpp
```diff
@@ -0,0 +1,123 @@
...
+        stmt_list->child[0] = (zend_ast*)block;
+    }
+}
+
+void aikido_ast_process(zend_ast *ast)
+{
+    insert_call_to_ast(ast);
+
+    if(original_ast_process){
+        original_ast_process(ast);
```
hansott, line 120: shouldn't happen that original is null? :D

hansott (commented):
🚀 awesome

lib/php-extension/HookAst.cpp
tudor-timcu: it can happen if there are weird sequences of hook / unhook (for example in Apache that does on the fly reload of plugins). So it's better to be safe than sorry :D

lib/php-extension/HandleShouldBlockRequest.cpp
tudor-timcu: If should_block is called after the user is set, it is still needed for user-based rate limiting, user blocking, etc...

lib/php-extension/Aikido.cpp
tudor-timcu: There's no issue in using it, it will be cached after the first use, so subsequent calls can't break stuff.

lib/php-extension/Aikido.cpp
tudor-timcu: That's right, need to move some things around. Will fix.

lib/php-extension/Action.cpp
tudor-timcu: Makes sense, added in the latest version.

lib/php-extension/Aikido.cpp
tudor-timcu: Should be fixed in the latest version.

lib/php-extension/HandleShouldBlockRequest.cpp
tudor-timcu: Also, we need to preserver should_block_request for people that are already added the middleware code in their apps, so it won't break with the new versions.

lib/php-extension/Aikido.cpp
```diff
@@ -45,6 +42,7 @@ PHP_MSHUTDOWN_FUNCTION(aikido) {
     /* If SAPI name is "cli" run in "simple" mode */
     if (AIKIDO_GLOBAL(sapi_name) == "cli") {
         AIKIDO_LOG_INFO("MSHUTDOWN finished earlier because we run in CLI mode!\n");
+        phpLifecycle.UnhookAll();
```
hansott, line 45: shouldn't we do the unhook always on module shutdown?  or add comment why we unhook here?

tudor-timcu: added comment

PR #195, opened 2025-05-27
Added monitoring-mode support for ips and user agents

tudor-timcu:


lib/request-processor/handle_blocking_request.go
```diff
@@ -58,14 +58,26 @@ func OnGetBlockingStatus() string {
...
+
+	if userAgentMonitored, userAgentMonitoredDescriptions := utils.IsUserAgentMonitored(userAgent); userAgentMonitored {
+		log.Infof("User Agent \"%s\" found in monitored lists: %v!", userAgent, userAgentMonitoredDescriptions)
+		go grpc.OnMonitoredUserAgentMatch(userAgentMonitoredDescriptions)
+	}
+
+	if userAgentBlocked, userAgentBlockedDescriptions := utils.IsUserAgentBlocked(userAgent); userAgentBlocked {
+		log.Infof("User Agent \"%s\" found in blocked lists: %v!", userAgent, userAgentBlockedDescriptions)
+		go grpc.OnMonitoredUserAgentMatch(userAgentBlockedDescriptions)
+		return GetStoreAction("blocked", "user-agent", userAgentBlockedDescriptions[0], userAgent)
```
hansott: If both the blocked and monitored match, we could do a single gRPC call?

lib/request-processor/handle_blocking_request.go
```diff
@@ -58,14 +58,26 @@ func OnGetBlockingStatus() string {
...
 	}
 
-	if userAgentBlocked, userAgentBlockedDescription := utils.IsUserAgentBlocked(userAgent); userAgentBlocked {
-		log.Infof("User Agent \"%s\" blocked due to: %s!", userAgent, userAgentBlockedDescription)
-		return GetStoreAction("blocked", "user-agent", userAgentBlockedDescription, userAgent)
+	if ipBlocked, ipBlockedDescriptions := utils.IsIpBlocked(ip); ipBlocked {
+		log.Infof("IP \"%s\" found in blocked lists: %v!", ip, ipBlockedDescriptions)
+		go grpc.OnMonitoredIpMatch(ipBlockedDescriptions)
+		return GetStoreAction("blocked", "ip", ipBlockedDescriptions[0], ip)
+	}
```
hansott, line 107: If both blocked and monitored match, we could do a single gRPC call?

tests/server/test_user_agent_blocked/test.py
```diff
@@ -11,20 +11,20 @@
...
         
-    response = php_server_get("/test")
+    response = php_server_get("/test", headers={"User-Agent": "1234googlebot1234"})
     assert_response_code_is(response, 200)
     assert_response_body_contains(response, "Something")
     
     apply_config("start_config.json")
         
-    response = php_server_get("/test")
+    response = php_server_get("/test", headers={"User-Agent": "1234googlebot1234"})
```
hansott, line 27: include test to see that monitored user agent is not blocked?

lib/request-processor/utils/utils.go
```diff
@@ -219,38 +221,68 @@ func IsUserBlocked(userID string) bool {
...
+	if blocklist.MatchString(userAgent) {
+		matchedDetails := []string{}
+		for key, valueRegex := range globals.CloudConfig.UserAgentDetails {
+			if valueRegex != nil && valueRegex.MatchString(userAgent) {
+				matchedDetails = append(matchedDetails, key)
+			}
+		}
+
+		if len(matchedDetails) == 0 {
+			matchedDetails = append(matchedDetails, "unknown")
```
hansott: do we need unknown here? I wouldn't report unknown in that case (we don't show it in the dashboard)

Could be interesting for debugging though

lib/request-processor/grpc/client.go
```diff
@@ -210,3 +210,35 @@ func OnMiddlewareInstalled() {
 	}
 	log.Debugf("OnMiddlewareInstalled sent via socket")
 }
+
+func OnMonitoredIpMatch(lists []string) {
```
hansott, line 214: if lists empty, early return?

lib/request-processor/grpc/client.go
```diff
@@ -210,3 +210,35 @@ func OnMiddlewareInstalled() {
...
+
+	_, err := client.OnMonitoredIpMatch(ctx, &protos.MonitoredIpMatch{Lists: lists})
+	if err != nil {
+		log.Warnf("Could not call OnMonitoredIpMatch")
+		return
+	}
+	log.Debugf("OnMonitoredIpMatch sent via socket")
+}
+
+func OnMonitoredUserAgentMatch(lists []string) {
```
hansott, line 230: if lists empty, early return?

lib/request-processor/handle_blocking_request.go
tudor-timcu: It's an async grpc call, cause it's only done for reporting. So it shouldn't be an issue if we do 2. Also, this should happen very rarely, that both IPs and UserAgents be matched.

lib/request-processor/handle_blocking_request.go
tudor-timcu: It's an async grpc call, cause it's only done for reporting. So it shouldn't be an issue if we do 2. Also, this should happen very rarely, that both IPs and UserAgents be matched.

lib/request-processor/utils/utils.go
tudor-timcu: If user agent details does not match one on one with the big pattern, I added a check so the list always contains something

lib/agent/grpc/request.go
```diff
@@ -26,6 +26,19 @@ func storeAttackStats(req *protos.AttackDetected) {
 	}
 }
 
+func storeMonitoredListsMatches(m *map[string]int, lists []string) {
+	if *m == nil {
+		*m = make(map[string]int)
```
hansott, line 31: maybe return? safer without modification?

hansott: We need to initialise `*m` so that we can modify it, wasn't clear to me at first.

tests/server/test_tor_monitored_ip/test.py
```diff
@@ -0,0 +1,30 @@
+import requests
+import time
+import sys
+from testlib import *
+
+'''
+1. Sets up the allowed IP address config for route '/test'. Checks that requests are blocked.
```
hansott:
```suggestion
1. Sets up the monitored IP addresses. Checks that requests are blocked.
```

PR #212, opened 2025-06-19
Invocation details

tudor-timcu:


lib/php-extension/Aikido.cpp
```diff
@@ -59,7 +59,7 @@ PHP_MSHUTDOWN_FUNCTION(aikido) {
 }
 
 PHP_RINIT_FUNCTION(aikido) {
-    ScopedTimer scopedTimer("request_init");
+    ScopedTimer scopedTimer("request_init", "request_op");
```
bitterpanda63, line 62: special op unique to PHP?

lib/php-extension/Environment.cpp
```diff
@@ -158,5 +158,5 @@ void LoadEnvironment() {
     AIKIDO_GLOBAL(token) = GetEnvString("AIKIDO_TOKEN", "");
     AIKIDO_GLOBAL(endpoint) = GetEnvString("AIKIDO_ENDPOINT", "https://guard.aikido.dev/");
     AIKIDO_GLOBAL(config_endpoint) = GetEnvString("AIKIDO_REALTIME_ENDPOINT", "https://runtime.aikido.dev/");
-    AIKIDO_GLOBAL(report_stats_interval_to_agent) = GetEnvNumber("AIKIDO_REPORT_STATS_INTERVAL", 10000);
+    AIKIDO_GLOBAL(report_stats_interval_to_agent) = GetEnvNumber("AIKIDO_REPORT_STATS_INTERVAL", 1000);
```
bitterpanda63, line 161: Why did it go from 10 to 1 second? is that related to the ops stuff?

lib/php-extension/include/Stats.h
```diff
@@ -3,20 +3,22 @@
...
+    std::string kind;
     std::chrono::high_resolution_clock::time_point start;
     uint64_t duration = 0;
     
    public:
     ScopedTimer();
-    ScopedTimer(std::string key);
-    void SetSink(std::string key);
+    ScopedTimer(std::string key, std::string kind);
+    void SetSink(std::string key, std::string kind = "");
```
bitterpanda63: why did we add this weird empty default? sink should never be empty

lib/php-extension/Environment.cpp
tudor-timcu: It's from 10.000 requests / php process to 1000 requests / php process.

lib/php-extension/Aikido.cpp
tudor-timcu: Yep, it's the PHP extension's main API

lib/php-extension/include/Stats.h
tudor-timcu: Make sense, fixed.

bitterpanda63 (approved):

PR #219, opened 2025-06-23
Fix path traversal filter bypass

PopoviciMarian:


lib/php-extension/HandlePathAccess.cpp
```diff
@@ -3,15 +3,15 @@
 /* Helper for handle pre file path access */
 void helper_handle_pre_file_path_access(char *filename, EVENT_ID &eventId) {
     if (strncmp(filename, "php://", 6) == 0 &&
```
tudor-timcu: Pls add commen: php it's only lowercase but /filter not

lib/request-processor/vulnerabilities/path-traversal/detectPathTraversal.go
```diff
@@ -1,6 +1,21 @@
...
+import (
+	"strings"
+)
+
+func extractResourceOrOriginal(filePath string) string {
+	// Convert to lowercase for case-insensitive comparison
+	lowerFilePath := strings.ToLower(filePath)
+	if strings.HasPrefix(lowerFilePath, "php://filter/") {
+		// Use original string for splitting to preserve case in the resource path
+		parts := strings.Split(filePath, "/resource=")
```
tudor-timcu: Pls add comment that resource is always lowercase (same as above)

tudor-timcu: maybe split only once.

PR #222, opened 2025-06-26
Enhance ParseFormData function to handle key-value pairs more robustl…

PopoviciMarian:
…y + Update GetBody to utilize GetPost for form data


lib/php-extension/Utils.cpp
```diff
@@ -61,4 +61,36 @@ std::string NormalizeAndDumpJson(const json& jsonObj) {
...
     // https://json.nlohmann.me/api/basic_json/dump/
     return jsonObj.dump(-1, ' ', false, json::error_handler_t::ignore);
+}
+
+std::string ArrayToJson(zval* array) {
+    if (!array) {
+        return "";
+    }
+
+    json query_json;
```
kapyteinaikido, line 71: What happens if there's invalid unicode?

See https://aikido-security.slack.com/archives/C0798MASFU4/p1738277669520659?thread_ts=1738275039.624149&cid=C0798MASFU4 for a different, though similar case.

PR #242, opened 2025-07-28
Upgrade zen-internals to 0.1.42

tudor-timcu:


.github/workflows/build.yml
```diff
@@ -188,7 +188,7 @@ jobs:
           echo $AIKIDO_VERSION
           echo "AIKIDO_VERSION=$AIKIDO_VERSION" >> $GITHUB_ENV
           echo "AIKIDO_LIBZEN=libzen_internals_${{ env.ARCH }}-unknown-linux-gnu.so" >> $GITHUB_ENV
-          echo "AIKIDO_LIBZEN_VERSION=0.1.37" >> $GITHUB_ENV
+          echo "AIKIDO_LIBZEN_VERSION=0.1.41" >> $GITHUB_ENV
```
hansott: be careful: we added a new return code for SQL tokenize errors

hansott: Looks good: https://github.com/AikidoSec/firewall-php/blob/main/lib/request-processor/vulnerabilities/sql-injection/detectSqlInjection.go#L43

tudor-timcu: Yep, looked it also. Should be fine

PR #272, opened 2025-09-17
Add handling for 302 redirect responses in request processor

PopoviciMarian:


lib/php-extension/HandleFileCompilation.cpp
```diff
@@ -1,6 +1,8 @@
 #include "Includes.h"
 
 zend_op_array* handle_file_compilation(zend_file_handle* file_handle, int type) {
+    EventCache old_eventCache;
+    old_eventCache.Copy(eventCache);
```
tudor-timcu: this should be a part of request cache, not event cache

lib/request-processor/context/event_getters.go
```diff
@@ -19,6 +19,15 @@ func GetOutgoingRequestResolvedIp() string {
 	return Context.Callback(C.OUTGOING_REQUEST_RESOLVED_IP)
 }
 
+func GetOutgoingRequestResponseCode() string {
```
tudor-timcu: Think we should do this mostly in the C++ layer. Something like this (HandleUrls.cpp - handle_post_curl_exec)
    ```
eventCache.outgoingRequestResponseCode = CallPhpFunctionCurlGetInfo(curlHandle, CURLINFO_RESPONSE_CODE);

    if (eventCache.outgoingRequestResponseCode == 300) { // 3**
        if (!requestCache.outgoingRequestUrl) {
            // first redirect in the chain
            requestCache.outgoingRequestUrl = eventCache.outgoingRequestUrl;
        } else {
            // subsequent redirect in the chain
            requestCache.outgoingRequestEffectiveUrl = eventCache.outgoingRequestUrl;
        }
    }
    else {
        if (requestCache.outgoingRequestUrl) {
            // we had a redirect chain -> we need to use the initial url and effective url (after redirects - last in the chain)
            eventCache.outgoingRequestUrl = requestCache.outgoingRequestUrl;
            eventCache.outgoingRequestEffectiveUrl = requestCache.outgoingRequestEffectiveUrl;
        }
        else {
            // no redirect chain -> we can use the url and effective url from the current request
            eventCache.outgoingRequestEffectiveUrl = CallPhpFunctionCurlGetInfo(curlHandle, CURLINFO_EFFECTIVE_URL);
        }

        // reset the url to empty string for the next request
        requestCache.outgoingRequestUrl = "";
    }
```

tudor-timcu (changes requested):

PR #275, opened 2025-09-21
Fix and add tests for GetIpFromRequest function

PopoviciMarian:


lib/request-processor/utils/utils.go
```diff
@@ -214,17 +214,56 @@ func IsIpBypassed(ip string) bool {
...
+				parts[i] = ip
+				continue
 			}
 		}
-		if isIP(ip) {
-			return ip
+
+		// IPv4 with port: "203.0.113.5:1234" -> "203.0.113.5"
+		if strings.Count(ip, ":") == 1 {
+			if host, _, err := net.SplitHostPort(ip); err == nil {
```
timokoessler: In theory this function would also support IPv6 with brackets?

timokoessler (approved):

lib/request-processor/utils/utils.go
PopoviciMarian: yes, you're right, I'm fixing it now

tudor-timcu (approved):

PR #298, opened 2025-10-13
One Agent to rule them all

tudor-timcu:
- Added support for only one Aikido Agent that will be spawned for each machine
  - The request processor instances check on startup if an agent is already running and connect to that socket
  - Checking is done using the socket file on the disk and by checking the running PID
  - Fallbacks are in place if the PID is not running, but sockets still on disk
- The new Agent support any number of different servers being hosted on the same machine (virtual hosts, multiple php servers with different versions)
  - This works by having a global map in the Aikido Agent that stores each different server instance with it's correspoding data
  - On the first request of a new server, the request processor registers into the Agent by providing the token
  - The agent creates a new entry in the global map, spawns necessary goroutines for that server, initializes it's custom global data and it's live from now on
  - The agent won't even register that server if a token it's not provided, it will do so only after the first successful token registration by the request processor
- This also fixes various crashes and inconsistencies on the restart mechanisms of apache / php-fpm, as the agent is not unloaded anymore and just re-used across restarts.
- The agent is stopped only on uninstall
- The E2E tests were updated to use this exact same mechanism, each test has it's own mock token that registers into one Agent that all the 60+ tests use on a single machine
- This is more in line with industry standards of having only one Agent per machine and allows us to further optimize things in the future by having all the PHP-related servers report through only one agent

### DIFF
--- a/lib/agent/main.go
+++ b/lib/agent/main.go
@@ -32,8 +32,7 @@ func serversCleanupRoutine(_ *ServerData) {
 		now := utils.GetTime()
 		lastConnectionTime := atomic.LoadInt64(&server.LastConnectionTime)
 		if now-lastConnectionTime > constants.MinServerInactivityForCleanup {
-			// Server has been inactive
-			log.Infof(log.MainLogger, "Server has been inactive for more than 2 minutes, unregistering...")
+			log.Infof(log.MainLogger, "Server \"AIK_RUNTIME_***%s\" has been inactive for more than 2 minutes, unregistering...", utils.AnonymizeToken(token))
 			server_utils.Unregister(token)
 		}
 	}


lib/agent/grpc/domain.go
```diff
@@ -1,23 +1,23 @@
 package grpc
 
 import (
-	"main/globals"
+	"main/aikido_types"
 	"main/utils"
 )
 
-func storeDomain(domain string, port uint32) {
+func storeDomain(server *aikido_types.ServerData, domain string, port uint32) {
```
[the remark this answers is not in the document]
tudor-timcu: @AikidoSec feedback: server is checked before the call to storeDomain in func (s *GrpcServer) OnDomain(ctx context.Context, req *protos.Domain) (*emptypb.Empty, error)

lib/agent/globals/globals.go
```diff
@@ -5,76 +5,56 @@ import (
...
+	defer ServersMutex.Unlock()
+	Servers[token] = NewServerData()
+	return Servers[token]
+}
+
+func DeleteServer(token string) *ServerData {
+	ServersMutex.Lock()
+	defer ServersMutex.Unlock()
+	delete(Servers, token)
+	return Servers[token]
```
hansott: This will always be `nil`?

package/rpm/aikido.spec
```diff
@@ -23,6 +23,13 @@ cp -rf opt/aikido-%{version}/* %{buildroot}/opt/aikido-%{version}
 
 echo "Starting the installation process for Aikido PHP Firewall v%{version}..."
 
+pids=$(pgrep -f aikido-agent)
+if [ -n "$pids" ]; then
+    echo "Stopping Aikido Agent processes: $pids"
+    kill -9 $pids
```
hansott, line 29: should we try a graceful kill first? or not important here? Since it's only a heartbeat now and then?

lib/agent/aikido_types/constants.go
```diff
@@ -16,7 +16,5 @@ const (
 	MaxAttackDetectedEventsPerInterval  = 100
 	AttackDetectedEventsIntervalInMs    = 60 * 60 * 1000 // 1 hour
 	MinStatsCollectedForRelevantMetrics = 1000
-	MaxNumberOfStoredUsers              = 2000
-	MaxNumberOfStoredRoutes             = 5000
-	MaxNumberOfStoredHostnames          = 2000
+	MinServerInactivityForCleanup       = 10 * 60 * 1000 // 10 minutes
```
willem-delbare: cleanup of what? let's comment

lib/agent/server_utils/server.go
```diff
@@ -0,0 +1,56 @@
...
+	server.AikidoConfig.LocalhostAllowedByDefault = req.GetLocalhostAllowedByDefault()
+	server.AikidoConfig.CollectApiSchema = req.GetCollectApiSchema()
+
+	log.SetLogLevel(server.AikidoConfig.LogLevel)
+	log.Init(server.AikidoConfig.DiskLogs)
+	log.Debugf("Updated Aikido Config with the one passed via gRPC!")
+}
+
+func Register(token string, req *protos.Config) {
+	log.Infof("Registering server %s...", utils.AnonymizeString(token))
```
willem-delbare: hmm why do we want this?

lib/agent/main.go
```diff
@@ -9,13 +9,35 @@ import (
...
+		server := globals.GetServer(token)
+		if server == nil {
+			continue
+		}
+		now := utils.GetTime()
+		lastConnectionTime := atomic.LoadInt64(&server.LastConnectionTime)
+		if now-lastConnectionTime > MinServerInactivityForCleanup {
+			// Server has been inactive for more than 10 minutes
+			log.Infof("Server has been inactive for more than 10 minutes, unregistering...")
+			server_utils.Unregister(token)
```
hansott, line 36: There's a brief chance that we might delete a just created server because `LastConnectionTime` is `0`  until `GetCloudConfig` is called? Should we set `LastConnectionTime` on creation of a server?

lib/agent/main.go
```diff
@@ -9,13 +9,35 @@ import (
...
 	"syscall"
+	"time"
 )
 
+var serversCleanupChannel = make(chan struct{})
+var serversCleanupTicker = time.NewTicker(10 * time.Minute)
+
+func serversCleanupRoutine(server *ServerData) {
+	for _, token := range globals.GetServersTokens() {
+		server := globals.GetServer(token)
```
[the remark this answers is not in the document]
hansott: seems a good one to fix? ^^

lib/agent/cloud/config.go
```diff
@@ -19,15 +18,15 @@ func CheckConfigUpdatedAt() {
 		return
 	}
 
-	if cloudConfigUpdatedAt.ConfigUpdatedAt <= globals.CloudConfig.ConfigUpdatedAt {
+	if cloudConfigUpdatedAt.ConfigUpdatedAt <= server.CloudConfig.ConfigUpdatedAt {
```
hansott: No CloudConfigMutex.Lock() needed here?

lib/php-extension/Agent.cpp
```diff
@@ -2,33 +2,83 @@
...
+    std::string aikidoRunFolder = "/var/run/aikido-" + std::string(PHP_AIKIDO_VERSION);
+    std::string socketPath = "";
+    
+    DIR* dir;
+    struct dirent* ent;
+    if ((dir = opendir(aikidoRunFolder.c_str())) != NULL) {
+        while ((ent = readdir(dir)) != NULL) {
+            std::string filename(ent->d_name);
+            if (filename.find(".sock") != std::string::npos) {
+                socketPath = aikidoRunFolder + "/" + filename;
```
hansott: Prob good idea to test connection? `ping` / `pong` on gRPC?

hansott: Do we need to grab latest one? readdir sorting might be random?

lib/agent/utils/utils.go
```diff
@@ -72,29 +72,29 @@ func StopPollingRoutine(stopChan chan struct{}) {
 	close(stopChan)
 }
```
hansott, line 72: any protection needed against closing channel again? panic?

Called in a few places...

lib/php-extension/Agent.cpp
```diff
@@ -2,33 +2,83 @@
...
+                }
+            }
+        }
+        closedir(dirTree);
+    }
+
+    return agentPID;
+}
+
+bool Agent::RemoveSocketFiles() {
```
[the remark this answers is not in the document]
hansott: Seems valid, see:

```cpp
if (!this->RemoveSocketFiles()) {
      AIKIDO_LOG_WARN("Failed to remove some socket files, will try to re-spawn Aikido Agent...\n");
  } else {
      AIKIDO_LOG_INFO("Successfully removed old socket files!\n");
  }
```

package/rpm/aikido.spec
tudor-timcu: we catch sigterm in agent, so it should die gracefully. but if it dies abruptly, it's still no issue, cause it's a standalone process

lib/agent/globals/globals.go
tudor-timcu: indeed, fixed

lib/agent/main.go
tudor-timcu: very good point, thanks for this

tudor-timcu: Fixed

lib/agent/server_utils/server.go
tudor-timcu: Changed this to anonymize token and printed only last 4 chars

lib/agent/cloud/config.go
tudor-timcu: very good point, fixed

lib/php-extension/Agent.cpp
tudor-timcu: in theory there should be only one there, but indeed we might end up with 2 sockets. Not sure how I would test the grpc connection over the socket, but will look to make this more robust.

lib/agent/utils/utils.go
tudor-timcu: by design, should be called only once per channel. some entry points have catches for the panics, but not all of them. whas thinking to add also  on the grpc ones.

lib/ipc.proto
```diff
@@ -16,59 +16,72 @@ service Aikido {
...
 message Config {
-	string token = 1;
-  string log_level = 2;
-  bool disk_logs = 3;
-  bool blocking = 4;
-  bool localhost_allowed_by_default = 5;
-  bool collect_api_schema = 6;
+  string platform_name = 1;
+  string platform_version = 2;
+	string token = 3;
```
hansott:
```suggestion
string token = 3;
```

lib/agent/grpc/request.go
```diff
@@ -254,31 +254,31 @@ func getRateLimitingDataForEndpoint(method, route, routeParsed string) *RateLimi
...
+	rateLimitingDataMatch := getRateLimitingDataForEndpoint(server, method, route, routeParsed)
 	if rateLimitingDataMatch == nil {
 		return &protos.RateLimitingStatus{Block: false}
 	}
 
 	if rateLimitGroup != "" {
 		// If the rate limit group exists, we only try to rate limit by rate limit group
 		if isRateLimitingThresholdExceeded(&rateLimitingDataMatch.Config, rateLimitingDataMatch.RateLimitGroupCounts, rateLimitGroup) {
-			log.Infof("Rate limited request for group %s - %s %s - %v", rateLimitGroup, method, routeParsed, rateLimitingDataMatch.RateLimitGroupCounts[rateLimitGroup])
+			log.Infof(server.Logger, "Rate limited request for group %s - %s %s - %v", rateLimitGroup, method, routeParsed, rateLimitingDataMatch.RateLimitGroupCounts[rateLimitGroup])
```
hansott, line 273: do we need to check if server is not `nil` here? At start of `GetRateLimitingStatus`?

lib/agent/grpc/request.go
```diff
@@ -294,30 +294,30 @@ func getIpsList(ipsList map[string]IpBlocklist) map[string]*protos.IpList {
 	return m
 }
 
-func getCloudConfig(configUpdatedAt int64) *protos.CloudConfig {
-	isBlockingEnabled := utils.IsBlockingEnabled()
+func getCloudConfig(server *ServerData, configUpdatedAt int64) *protos.CloudConfig {
+	isBlockingEnabled := utils.IsBlockingEnabled(server)
```
hansott, line 306: do we need to check if server is not `nil` here?

lib/agent/cloud/cloud.go
```diff
@@ -2,28 +2,19 @@ package cloud
...
+	utils.StartPollingRoutine(server.PollingData.HeartbeatRoutineChannel, server.PollingData.HeartbeatTicker, SendHeartbeatEvent, server)
+	utils.StartPollingRoutine(server.PollingData.ConfigPollingRoutineChannel, server.PollingData.ConfigPollingTicker, CheckConfigUpdatedAt, server)
 }
 
-func Uninit() {
-	utils.StopPollingRoutine(HeartbeatRoutineChannel)
-	utils.StopPollingRoutine(ConfigPollingRoutineChannel)
+func Uninit(server *ServerData) {
+	utils.StopPollingRoutine(server.PollingData.HeartbeatRoutineChannel)
+	utils.StopPollingRoutine(server.PollingData.ConfigPollingRoutineChannel)
```
hansott, line 19: Do we need to Stop the tickers here?

lib/agent/grpc/request.go
tudor-timcu: fixed, it was checked above on the callstack, but better safe than sorry :D

lib/agent/grpc/request.go
tudor-timcu: fixed

lib/agent/cloud/cloud.go
tudor-timcu: tickers are stopped automatically when the channel is closed

lib/agent/main.go
```diff
@@ -2,70 +2,102 @@ package main
...
+		if server == nil {
+			continue
+		}
+		now := utils.GetTime()
+		lastConnectionTime := atomic.LoadInt64(&server.LastConnectionTime)
+		if now-lastConnectionTime > constants.MinServerInactivityForCleanup {
+			// Server has been inactive
+			log.Infof(log.MainLogger, "Server has been inactive for more than 2 minutes, unregistering...")
```
hansott: maybe worth including the token here?

lib/request-processor/vulnerabilities/ssrf/containsPrivateIPAddress_test.go
```diff
@@ -179,6 +179,7 @@ func TestPublicIPs(t *testing.T) {
 }
 
 // TestPrivateIPs checks that private IPs return true for private IP check.
+/*
```
hansott, line 182: this is intended?

lib/agent/main.go
tudor-timcu: Done

lib/request-processor/vulnerabilities/ssrf/containsPrivateIPAddress_test.go
tudor-timcu: yeah, does not pass in a new go packages version. this was solved by Marian on a different branch.

PR #324, opened 2025-10-29
Fixed request processing bottleneck by updating config faster and only when needed

tudor-timcu:
- For php-fpm, reload the config only once per process, as php-fpm can serve only one site per process, so additional reloads will just slow down processing
- Apache (mod-php) behaves in a different way - it can server any site from any process, so the token can be changed for every request. Added a simplified function that only refreshes the token, not the whole config and this makes it faster for Apache also, even though we need to refresh token every time.
- The issue on fly.io was that this introduced a slowdown for request processing, so when a lot of requests were coming in, the queue of requests in php-fpm grew faster than we could process these (also 512mb ram was contributing to this overall slowdown) - with these fixes it's good (deployed it on fly)
- Some other small fixes: make sure we kill all agent pids in inconsistent states

### DIFF
--- a/lib/php-extension/RequestProcessor.cpp
+++ b/lib/php-extension/RequestProcessor.cpp
@@ -207,7 +207,7 @@ void RequestProcessor::LoadConfigFromEnvironment() {
     LoadConfig(previousToken, currentToken);
 }
 
-void RequestProcessor::LoadConfigWithTokenFromMiddleware(const std::string& tokenFromMiddleware) {
+void RequestProcessor::LoadConfigWithTokenFromPHPSetToken(const std::string& tokenFromMiddleware) {
     LoadConfig(AIKIDO_GLOBAL(token), tokenFromMiddleware);
 }
 
--- a/lib/php-extension/HandleSetToken.cpp
+++ b/lib/php-extension/HandleSetToken.cpp
@@ -19,6 +19,6 @@ ZEND_FUNCTION(set_token) {
         RETURN_BOOL(false);
     }
 
-    requestProcessor.LoadConfigWithTokenFromMiddleware(std::string(token, tokenLength));
+    requestProcessor.LoadConfigWithTokenFromPHPSetToken(std::string(token, tokenLength));
     RETURN_BOOL(true);
 }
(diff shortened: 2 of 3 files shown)


lib/php-extension/RequestProcessor.cpp
```diff
@@ -160,7 +166,20 @@ bool RequestProcessor::RequestInit() {
         return false;
     }
     
-    this->LoadConfig();
+    if (AIKIDO_GLOBAL(sapi_name) == "apache2handler") {
+        // Apache-mod-php can serve multiple sites per process, so the config should be loaded for each request,
+        // as the token can be different for each request.
```
hansott:
```suggestion
      // Apache-mod-php can serve multiple sites per process
      // We need to reload config each request to detect token changes
      // (either via set_token() API or if AIKIDO_TOKEN environment variable changed between requests)
```

lib/php-extension/RequestProcessor.cpp
```diff
@@ -187,11 +197,22 @@ void RequestProcessor::LoadConfig(std::string userProvidedToken) {
...
 }
 
+void RequestProcessor::LoadConfigFromEnvironment() {
+    std::string previousToken = AIKIDO_GLOBAL(token);
+    LoadEnvironment();
+    std::string currentToken = AIKIDO_GLOBAL(token);
+    LoadConfig(previousToken, currentToken);
+}
+
+void RequestProcessor::LoadConfigWithTokenFromMiddleware(const std::string& tokenFromMiddleware) {
```
hansott:
```suggestion
void RequestProcessor::LoadConfigWithTokenFromPHPSetToken(const std::string& tokenFromMiddleware) {
```

Wouldn't this reflect the name better for `set_token(...)` usage?

lib/php-extension/RequestProcessor.cpp
```diff
@@ -160,7 +159,21 @@ bool RequestProcessor::RequestInit() {
         return false;
     }
     
-    this->LoadConfig();
+    if (AIKIDO_GLOBAL(sapi_name) == "apache2handler") {
+      // Apache-mod-php can serve multiple sites per process
+      // We need to reload config each request to detect token changes
+      // (either via set_token() API or if AIKIDO_TOKEN environment variable changed between requests)
```
hansott:
```suggestion
```

No longer relevant than?

PR #355, opened 2025-11-18
Added support for ZTS (Zend Thread Safety), including FrankenPHP support

ioaniftimesei:
- Updated RequestProcessor to manage instances with thread safety, supporting both NTS and ZTS modes.
- Introduced CreateInstance and DestroyInstance functions for instance management.
- Refactored existing function pointers to accept instance pointers, ensuring proper context handling.
- Improved logging and error handling during initialization and configuration updates.
- Added mutex locks to prevent race conditions in concurrent environments.
- Adjusted various function signatures to accommodate instance pointers for better encapsulation of state.

## Summary by Aikido
| ⚠️ Security Issues: 7 |  Quality Issues: 0 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Added ZTS thread-safety with instance-based RequestProcessor support.

**⚡ Enhancements**
* Introduced per-instance init and config reload logic for tokens.
* Added mutexes and synchronization to protect shared server and rate-limiting state.
* Integrated FrankenPHP test modes and improved test runner startup logic.
* Exposed thread utilities and added threading headers to PHP extension.

**🔧 Refactors**
* Refactored many function signatures to accept instance pointers.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/84897838?groupId=6)</sup>

### DIFF
--- a/lib/php-extension/RequestProcessor.cpp
+++ b/lib/php-extension/RequestProcessor.cpp
@@ -111,9 +111,6 @@ void RequestProcessor::Uninit() {
     this->libHandle = nullptr;
     AIKIDO_LOG_INFO("Aikido Request Processor unloaded!\n");
 }
-RequestProcessor::~RequestProcessor() {
-    this->Uninit();
-}
 
 bool RequestProcessorInstance::ContextInit() {
     if (!this->requestInitialized || requestProcessor.requestProcessorContextInitFn == nullptr || this->requestProcessorInstance == nullptr) {
@@ -324,4 +321,5 @@ RequestProcessorInstance::~RequestProcessorInstance() {
         requestProcessor.destroyInstanceFn(this->threadId);
         this->requestProcessorInstance = nullptr;
     }
+    requestProcessor.Uninit();
 }
--- a/lib/php-extension/include/RequestProcessor.h
+++ b/lib/php-extension/include/RequestProcessor.h
@@ -34,7 +34,7 @@ class RequestProcessor {
     RequestProcessorUninitFn requestProcessorUninitFn = nullptr;
 
     RequestProcessor() = default;
-    ~RequestProcessor();
+    ~RequestProcessor() = default;
 
     std::string GetInitData(const std::string& userProvidedToken = "");
 


lib/request-processor/main.go
```diff
@@ -17,17 +18,39 @@ import (
...
-	C.EVENT_SET_USER:                 OnUserEvent,
-	C.EVENT_SET_RATE_LIMIT_GROUP:     OnRateLimitGroupEvent,
-	C.EVENT_GET_AUTO_BLOCKING_STATUS: OnGetAutoBlockingStatus,
-	C.EVENT_GET_BLOCKING_STATUS:      OnGetBlockingStatus,
-	C.EVENT_PRE_OUTGOING_REQUEST:     OnPreOutgoingRequest,
-	C.EVENT_POST_OUTGOING_REQUEST:    OnPostOutgoingRequest,
-	C.EVENT_PRE_SHELL_EXECUTED:       OnPreShellExecuted,
-	C.EVENT_PRE_PATH_ACCESSED:        OnPrePathAccessed,
-	C.EVENT_PRE_SQL_QUERY_EXECUTED:   OnPreSqlQueryExecuted,
+	C.EVENT_PRE_REQUEST: func(i interface{}) string {
```
tudor-timcu: This should be refactored to look like before

lib/agent/grpc/request.go
```diff
@@ -176,20 +176,6 @@ func incrementSlidingWindowEntry(m map[string]*SlidingWindow, key string) *Slidi
 	return entry
 }
 
-func updateRateLimitingCounts(server *ServerData, method string, route string, routeParsed string, user string, ip string, rateLimitGroup string) {
```
tudor-timcu, line 179: This was done separately (and not inside getRateLimitingStatus) because it is run async from OnRequestShutdown, while getRateLimitingStatus runs blocking (the request processor wait for the rate limiting result). getRateLimitingStatus should do the miminal number of operations. I think we should go back to the way this was before.

lib/agent/aikido_types/stats.go
```diff
@@ -48,6 +48,7 @@ type RateLimitingValue struct {
 	UserCounts           map[string]*SlidingWindow
 	IpCounts             map[string]*SlidingWindow
 	RateLimitGroupCounts map[string]*SlidingWindow
+	Mutex                sync.Mutex
```
tudor-timcu, line 51: If we go back to the initial logic, this should be removed.

lib/php-extension/Handle.cpp
```diff
@@ -36,6 +40,7 @@ ZEND_NAMED_FUNCTION(aikido_generic_handler) {
     std::string outputEvent;
     bool caughtException = false;
 
+    auto& eventCache = AIKIDO_GLOBAL(eventCache);
```
tudor-timcu: I like that we get the reference to these global vars to not go through TLS each time.
I would prefer if we ditch auto though, makes the code more readable in my opinion.

lib/php-extension/PhpWrappers.cpp
```diff
@@ -36,12 +35,16 @@ bool CallPhpFunction(std::string function_name, unsigned int params_number, zval
 
     int _result = call_user_function(EG(function_table), object, &_function_name, _return_value, params_number, params);
 
-    zend_string_release(_function_name_str);
+    zval_dtor(&_function_name);
+
+    if (_result != SUCCESS) {
```
tudor-timcu: If this was moved up, aren't we skipping the cleanup for _temp_return_value in some cases?
Also, did we check that zval_dtor also clears the function_name string?

lib/php-extension/RequestProcessor.cpp
```diff
@@ -1,18 +1,17 @@
 #include "Includes.h"
 
-RequestProcessor requestProcessor;
-
 std::string RequestProcessor::GetInitData(const std::string& token) {
```
tudor-timcu: Let's rename the parameter as userProvidedToken

lib/php-extension/RequestProcessor.cpp
```diff
@@ -173,43 +187,77 @@ bool RequestProcessor::RequestInit() {
...
     
-    if (AIKIDO_GLOBAL(sapi_name) == "apache2handler") {
-      // Apache-mod-php can serve multiple sites per process
+    const auto& sapiName = AIKIDO_GLOBAL(sapi_name);
+    if (sapiName == "apache2handler" || sapiName == "frankenphp") {
+      // Apache-mod-php and FrankenPHP can serve multiple sites per process
       // We need to reload config each request to detect token changes
+      // Check disable BEFORE modifying any state (shared Go state or per-instance state)
+      // Use GetEnvBool() to read disable flag without modifying global state
+        if (GetEnvBool("AIKIDO_DISABLE", false)) {
```
tudor-timcu: Don't think this should be here, as the config was not yet reloaded and AIKIDO_DISABLE could change after that.

lib/php-extension/RequestProcessor.cpp
```diff
@@ -173,43 +187,77 @@ bool RequestProcessor::RequestInit() {
...
         this->LoadConfigFromEnvironment();
     } else {
-        // Server APIs that are not apache-mod-php (like php-fpm, cli-server, ...) 
+        // Server APIs that are not apache-mod-php/frankenphp (like php-fpm, cli-server, ...) 
         //  can only serve one site per process, so the config should be loaded at the first request.
         // If the token is not set at the first request, we try to reload it until we get a valid token.
         // The user can update .env file via zero downtime deployments after the PHP server is started.
         if (AIKIDO_GLOBAL(token) == "") {
             AIKIDO_LOG_INFO("Loading Aikido config until we get a valid token for SAPI: %s...\n", AIKIDO_GLOBAL(sapi_name).c_str());
+            if (GetEnvBool("AIKIDO_DISABLE", false)) {
```
tudor-timcu: Same as above.

lib/request-processor/context/event_context.go
```diff
@@ -15,10 +17,22 @@ type EventContextData struct {
...
 }
 
-var EventContext EventContextData
+func getEventContext(inst *instance.RequestProcessorInstance) *EventContextData {
+	if inst == nil {
+		return nil
+	}
+	tid := inst.GetThreadID()
+	// Create new event context if it doesn't exist
+	newCtx := &EventContextData{}
```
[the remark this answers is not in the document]
tudor-timcu: +1

lib/request-processor/context/request_context.go
```diff
@@ -44,143 +48,198 @@ type RequestContextData struct {
...
+
 	return true
 }
 
-func Clear() bool {
-	Context = RequestContextData{
-		Callback: Context.Callback,
+func GetContext(inst *instance.RequestProcessorInstance) *RequestContextData {
+	tid := inst.GetThreadID()
+	return globals.GetFromThreadStorage[*RequestContextData](tid, &globals.ContextData)
```
tudor-timcu: Don't we pass around the instance pointer created from C++? Why do we still need the global thread storage that we access each time when GetContext is called? This add performance overhead and think we can do it differently.

tudor-timcu (changes requested):

lib/php-extension/HandleBypassedIp.cpp
```diff
@@ -1,26 +1,35 @@
...
+        AIKIDO_GLOBAL(requestProcessor).SendEvent(EVENT_GET_IS_IP_BYPASSED, output);
+        AIKIDO_GLOBAL(action).Execute(output);
     } catch (const std::exception &e) {
         AIKIDO_LOG_ERROR("Exception encountered in processing IP bypass check event: %s\n", e.what());
     }
 }
 
-
 bool IsAikidoDisabledOrBypassed() {
-    return AIKIDO_GLOBAL(disable) == true || isIpBypassed;
```
tudor-timcu, line 24: See if it's ok here @PopoviciMarian. There 2 variables, before was only one. (seems like a lazy init check).

lib/php-extension/PhpWrappers.cpp
```diff
@@ -17,7 +17,6 @@ bool CallPhpEcho(std::string message) {
 
 bool CallPhpFunction(std::string function_name, unsigned int params_number, zval* params, zval* return_value, zval* object) {
     if (!object && !zend_hash_str_exists(CG(function_table), function_name.c_str(), function_name.size())) {
-        AIKIDO_LOG_INFO("Function name '%s' does not exist!\n", function_name.c_str());
```
tudor-timcu, line 20: Revert this please

tudor-timcu (changes requested):

lib/agent/aikido_types/stats.go
tudor-timcu: Why do we need this mutex

lib/agent/cloud/cloud.go
```diff
@@ -11,8 +11,8 @@ func Init(server *ServerData) {
 
 	CheckConfigUpdatedAt(server)
 
-	utils.StartPollingRoutine(server.PollingData.HeartbeatRoutineChannel, server.PollingData.HeartbeatTicker, SendHeartbeatEvent, server)
 	utils.StartPollingRoutine(server.PollingData.ConfigPollingRoutineChannel, server.PollingData.ConfigPollingTicker, CheckConfigUpdatedAt, server)
+	utils.StartPollingRoutine(server.PollingData.HeartbeatRoutineChannel, server.PollingData.HeartbeatTicker, SendHeartbeatEvent, server)
```
tudor-timcu, line 14: these should be as before

lib/agent/grpc/request.go
```diff
@@ -321,31 +307,38 @@ func getRateLimitingStatus(server *ServerData, method, route, routeParsed, user,
...
 		return &protos.RateLimitingStatus{Block: false}
 	}
 
+	rateLimitingDataMatch.Mutex.Lock()
+	server.RateLimitingMutex.RUnlock()
+	defer rateLimitingDataMatch.Mutex.Unlock()
+
 	if rateLimitGroup != "" {
 		// If the rate limit group exists, we only try to rate limit by rate limit group
 		if isRateLimitingThresholdExceeded(&rateLimitingDataMatch.Config, rateLimitingDataMatch.RateLimitGroupCounts, rateLimitGroup) {
```
tudor-timcu: We should create a new function:
- isRateLimitingThresholdExceededAndIncrement
In this function we should lock rateLimitingDataMatch and do these common operations

lib/agent/grpc/server.go
```diff
@@ -76,15 +91,14 @@ func (s *GrpcServer) OnRequestShutdown(ctx context.Context, req *protos.RequestM
...
+
 	log.Debugf(server.Logger, "Received request metadata: %s %s %d %s %s %v", req.GetMethod(), req.GetRouteParsed(), req.GetStatusCode(), req.GetUser(), req.GetIp(), req.GetApiSpec())
 	if req.GetShouldDiscoverRoute() || req.GetRateLimited() {
 		go storeTotalStats(server, req.GetRateLimited())
 		go storeRoute(server, req.GetMethod(), req.GetRouteParsed(), req.GetApiSpec(), req.GetRateLimited())
-		go updateRateLimitingCounts(server, req.GetMethod(), req.GetRoute(), req.GetRouteParsed(), req.GetUser(), req.GetIp(), req.GetRateLimitGroup())
 	}
 	go updateAttackWaveCountsAndDetect(server, req.GetIsWebScanner(), req.GetIp(), req.GetUser(), req.GetUserAgent(), req.GetMethod(), req.GetUrl())
 
-	atomic.StoreUint32(&server.GotTraffic, 1)
```
tudor-timcu, line 87: Why is this removed?

lib/agent/grpc/server.go
```diff
@@ -162,6 +182,7 @@ func (s *GrpcServer) OnMonitoredUserAgentMatch(ctx context.Context, req *protos.
 	if server == nil {
 		return &emptypb.Empty{}, nil
 	}
+
```
tudor-timcu: Can we remove extra new lines, so the diff is clearer.

lib/agent/rate_limiting/rate_limiting.go
```diff
@@ -6,19 +6,24 @@ import (
...
-
+	server.RateLimitingMutex.RLock()
+	endpoints := make([]*RateLimitingValue, 0, len(server.RateLimitingMap))
 	for _, endpoint := range server.RateLimitingMap {
+		endpoints = append(endpoints, endpoint)
+	}
+	server.RateLimitingMutex.RUnlock()
+
+	for _, endpoint := range endpoints {
+		endpoint.Mutex.Lock()
```
tudor-timcu, line 13: we should use defer endpoint.Mutex.Unlock() here

lib/agent/server_utils/server.go
```diff
@@ -28,26 +28,38 @@ func storeConfig(server *ServerData, req *protos.Config) {
 	server.AikidoConfig.CollectApiSchema = req.GetCollectApiSchema()
 }
 
-func Register(serverKey ServerKey, requestProcessorPID int32, req *protos.Config) {
```
tudor-timcu, line 31: I don't understand why we removed register entirely and why are all of these new changes in server.go needed.

tudor-timcu: This file should be mostly as before, we should only fix the race condition.

lib/agent/server_utils/server.go
```diff
@@ -28,26 +28,38 @@ func storeConfig(server *ServerData, req *protos.Config) {
...
+	atomic.StoreInt64(&server.LastConnectionTime, utils.GetTime())
+}
 
+func CompleteServerConfiguration(server *ServerData, serverKey ServerKey, req *protos.Config) {
 	log.InfofMainAndServer(server.Logger, "Server \"AIK_RUNTIME_***%s\" (server PID: %d) registered successfully!", utils.AnonymizeToken(serverKey.Token), serverKey.ServerPID)
 
-	atomic.StoreInt64(&server.LastConnectionTime, utils.GetTime())
-
 	cloud.Init(server)
+	rate_limiting.StartRateLimitingTicker(server)
```
tudor-timcu: these should be named as before.

lib/agent/aikido_types/init_data.go
```diff
@@ -215,6 +215,10 @@ type ServerData struct {
 	// Got some request info passed via gRPC to the Agent
 	GotTraffic uint32
 
+	// Tracks if the "started" event has been sent for this server
+	// In multi-worker mode (e.g., frankenphp-worker), only one worker should send it
+	SentStartedEvent uint32
```
tudor-timcu: this should also be reverted.

lib/agent/attack-wave-detection/attackWaveDetector.go
```diff
@@ -19,9 +19,9 @@ func AdvanceAttackWaveQueues(server *ServerData) {
 	}
 }
 
-func Init(server *ServerData) {
+// StartAttackWaveTicker starts the attack wave detection ticker
+func StartAttackWaveTicker(server *ServerData) {
 	utils.StartPollingRoutine(server.PollingData.AttackWaveChannel, server.PollingData.AttackWaveTicker, AdvanceAttackWaveQueues, server)
-	AdvanceAttackWaveQueues(server)
```
tudor-timcu, line 24: this should be reverted

lib/agent/cloud/event_started.go
```diff
@@ -4,9 +4,16 @@ import (
 	. "main/aikido_types"
 	"main/constants"
 	"main/utils"
+	"sync/atomic"
 )
 
 func SendStartEvent(server *ServerData) {
+	// In multi-worker mode (e.g., frankenphp-worker), ensure only one worker sends the started event
+	// Use atomic compare-and-swap to guarantee exactly-once semantics
+	if !atomic.CompareAndSwapUint32(&server.SentStartedEvent, 0, 1) {
```
tudor-timcu: this should be reverted.

lib/agent/aikido_types/stats.go
tudor-timcu: I think this is good.

lib/php-extension/Action.cpp
```diff
@@ -1,12 +1,12 @@
...
-Action action;
-
 ACTION_STATUS Action::executeThrow(json &event) {
     int _code = event["code"].get<int>();
     std::string _message = event["message"].get<std::string>();
-    zend_throw_exception(GetFirewallDefaultExceptionCe(), _message.c_str(), _code);
+    
     CallPhpFunctionWithOneParam("http_response_code", _code);
+    zend_throw_exception(GetFirewallDefaultExceptionCe(), _message.c_str(), _code);
+
```
tudor-timcu, line 13: Can we add a comment here

lib/php-extension/Aikido.cpp
```diff
@@ -52,35 +59,93 @@ PHP_MSHUTDOWN_FUNCTION(aikido) {
...
 }
 
-PHP_RSHUTDOWN_FUNCTION(aikido) {
+// Common request shutdown logic shared between RSHUTDOWN and aikidoworker_rshutdown
+static void aikido_do_request_shutdown() {
     ScopedTimer scopedTimer("request_shutdown", "request_op");
 
     AIKIDO_LOG_DEBUG("RSHUTDOWN started!\n");
 
-    if (AIKIDO_GLOBAL(disable) == true) {
```
tudor-timcu, line 73: let's revert. it's the same thing.

lib/php-extension/HookAst.cpp
```diff
@@ -1,8 +1,5 @@
 #include "Includes.h"
```
tudor-timcu, line 1: revert renames global_ast_to_clean -> globalAstToClean

lib/php-extension/RequestProcessor.cpp
```diff
@@ -156,11 +174,7 @@ bool RequestProcessor::Init() {
         return false;
     }
 
-    if (!requestProcessorInitFn(GoCreateString(initDataString))) {
-        AIKIDO_LOG_ERROR("Failed to initialize Aikido Request Processor library: %s!\n", dlerror());
-        this->initFailed = true;
-        return false;
-    }
+    this->requestProcessorInitFn = requestProcessorInitFn;
```
tudor-timcu: move this in line with other fn initialization upper

tudor-timcu (changes requested):

lib/request-processor/instance/wrapper.go
```diff
@@ -0,0 +1,157 @@
...
+// In NTS mode (standard PHP), there's one global instance.
+// In ZTS mode (FrankenPHP), each thread gets its own instance with locking.
+type RequestProcessorInstance struct {
+	CurrentToken    string
+	CurrentServer   *ServerData
+	threadID        uint64         // CACHED: OS thread ID cached at RINIT for fast context lookups
+	ContextInstance unsafe.Pointer // For context callbacks
+	ContextCallback unsafe.Pointer // C function pointer, must be per-instance in ZTS
+
+	RequestContext interface{}
```
tudor-timcu, line 21: add comment why this is interface{}

lib/agent/grpc/request.go
```diff
@@ -315,34 +304,48 @@ func getRateLimitingDataForEndpoint(server *ServerData, method, route, routePars
...
 }
 
+func isRateLimitingThresholdExceededAndIncrement(rateLimitingDataMatch *RateLimitingValue,
+	countsMap map[string]*SlidingWindow,
+	key string,
+) bool {
+	rateLimitingDataMatch.Mutex.Lock()
+	defer rateLimitingDataMatch.Mutex.Unlock()
+
+	if isRateLimitingThresholdExceeded(&rateLimitingDataMatch.Config, countsMap, key) {
```
tudor-timcu, line 314:
```suggestion
	isRateLimited := isRateLimitingThresholdExceeded(&rateLimitingDataMatch.Config, countsMap, key)
	incrementSlidingWindowEntry(countsMap, key)

	return isRateLimited
```

lib/agent/rate_limiting/rate_limiting.go
```diff
@@ -6,19 +6,24 @@ import (
 )
 
 func AdvanceRateLimitingQueues(server *ServerData) {
-	server.RateLimitingMutex.Lock()
-	defer server.RateLimitingMutex.Unlock()
-
+	server.RateLimitingMutex.RLock()
+	defer server.RateLimitingMutex.RUnlock()
+	endpoints := make([]*RateLimitingValue, 0, len(server.RateLimitingMap))
```
tudor-timcu: why do we make this temp array? it should work fine with the previous for loop

lib/agent/rate_limiting/rate_limiting.go
tudor-timcu: this still needs to be fixed

lib/agent/rate_limiting/rate_limiting.go
```diff
@@ -6,19 +6,24 @@ import (
...
 		AdvanceSlidingWindowMap(endpoint.UserCounts, endpoint.Config.WindowSizeInMinutes)
 		AdvanceSlidingWindowMap(endpoint.IpCounts, endpoint.Config.WindowSizeInMinutes)
 		AdvanceSlidingWindowMap(endpoint.RateLimitGroupCounts, endpoint.Config.WindowSizeInMinutes)
+		endpoint.Mutex.Unlock()
 	}
 }
 
 func Init(server *ServerData) {
 	utils.StartPollingRoutine(server.PollingData.RateLimitingChannel, server.PollingData.RateLimitingTicker, AdvanceRateLimitingQueues, server)
-	AdvanceRateLimitingQueues(server)
```
tudor-timcu, line 21: why is this removed from Init?

lib/agent/server_utils/server.go
```diff
@@ -29,9 +29,16 @@ func storeConfig(server *ServerData, req *protos.Config) {
 }
 
 func Register(serverKey ServerKey, requestProcessorPID int32, req *protos.Config) {
+
+	server := globals.GetServer(serverKey)
```
tudor-timcu: there's still room for race codition like this also.

tudor-timcu: This function should be renamed to RegisterServerIfNotExists or smth like that. It should do only one write lock with defer at the beginning of the function and check if the server exists. if not, it should register it.

lib/php-extension/HookAst.cpp
tudor-timcu: this is not fixed

lib/request-processor/aikido_types/server.go
```diff
@@ -10,5 +10,7 @@ type ServerData struct {
 	CloudConfig         CloudConfigData
 	CloudConfigMutex    sync.Mutex
 	MiddlewareInstalled bool
+	ServerInitialized   bool
```
tudor-timcu, line 13: we should use atomic operations here instead of mutex

tudor-timcu: Also, not really sure we need this. Can't we leverage ServersMutex and just lookup if the server is in there?

lib/request-processor/config/config.go
```diff
@@ -5,19 +5,22 @@ import (
...
 	"main/log"
 	"main/utils"
 	"os"
 )
 
-func UpdateToken(token string) bool {
-	if token == globals.CurrentToken {
-		log.Debugf("Token is the same as previous one, skipping config reload...")
+func UpdateToken(instance *instance.RequestProcessorInstance, token string) bool {
+	server := globals.GetServer(token)
```
tudor-timcu, line 19: let's move this down:
```suggestion
	instance.SetCurrentServer(globals.GetServer(token))
```

lib/request-processor/grpc/client.go
```diff
@@ -189,24 +206,28 @@ func GetCloudConfigForAllServers(timeout time.Duration) {
 	}
 }
 
-func OnUserEvent(server *ServerData, id string, username string, ip string) {
+func OnUserEvent(threadID uint64, server *ServerData, token string, id string, username string, ip string) {
```
tudor-timcu: we really need to pass in the threadID just for logging? isn't there a nicer way to do this?

lib/request-processor/grpc/config.go
```diff
@@ -144,6 +147,14 @@ func setCloudConfig(server *ServerData, cloudConfigFromAgent *protos.CloudConfig
 }
 
 func StartCloudConfigRoutine() {
+	cloudConfigMutex.Lock()
+	defer cloudConfigMutex.Unlock()
+
+	if cloudConfigStarted {
```
tudor-timcu: this cloudConfigStarted and cloudConfigMutex variables should not be needed. at best, use atomic for this

package/rpm/aikido.spec
```diff
@@ -117,6 +143,36 @@ for PHP_VERSION in "${PHP_VERSIONS[@]}"; do
     fi
 done
 
+if [ -n "$FRANKENPHP_PHP_VERSION" ]; then
```
tudor-timcu, line 146: there's only 1 frankenphp version that can be installed at a time?

tudor-timcu (changes requested):

lib/agent/server_utils/server.go
```diff
@@ -29,9 +29,13 @@ func storeConfig(server *ServerData, req *protos.Config) {
...
+	server, exists := globals.GetOrCreateServer(serverKey)
+	if exists {
+		log.Debugf(log.MainLogger, "Server \"AIK_RUNTIME_***%s\" already exists, skipping registration (request processor PID: %d, server PID: %d)", utils.AnonymizeToken(serverKey.Token), requestProcessorPID, serverKey.ServerPID)
+		return
+	}
 	log.Infof(log.MainLogger, "Client (request processor PID: %d) connected. Registering server \"AIK_RUNTIME_***%s\" (server PID: %d)...", requestProcessorPID, utils.AnonymizeToken(serverKey.Token), serverKey.ServerPID)
 
-	server := globals.CreateServer(serverKey)
 	storeConfig(server, req)
 	server.Logger = log.CreateLogger(utils.AnonymizeToken(serverKey.Token), server.AikidoConfig.LogLevel, server.AikidoConfig.DiskLogs)
```
tudor-timcu, line 43: we should lock this whole function to avoid race condition:
	ServersMutex.Lock()
	defer ServersMutex.Unlock()

lib/php-extension/RequestProcessor.cpp
```diff
@@ -169,17 +183,53 @@ bool RequestProcessor::Init() {
...
+        }
+    }
+
     if (!this->Init()) {
         AIKIDO_LOG_ERROR("Failed to initialize the request processor: %s!\n", dlerror());
         return false;
     }
+    if (this->requestProcessorInstance == nullptr && this->createInstanceFn != nullptr) {
+        this->threadId = GetThreadID();
+        #ifdef ZTS
```
tudor-timcu:
```suggestion
        bool isZTS = (ZTS == 1);
```

lib/request-processor/config/config.go
```diff
@@ -45,18 +48,18 @@ func ReloadAikidoConfig(conf *AikidoConfigData, initJson string) ReloadResult {
 	}
 
 	if globals.ServerExists(conf.Token) {
-		if !UpdateToken(conf.Token) {
+		if !UpdateToken(instance, conf.Token) {
 			return ReloadWithSameToken
 		}
 		return ReloadWithPastSeenToken
 	}
```
tudor-timcu, line 58: We should add a comment here for FrankenPHP to explain that CreateServer would never be executed for FrankenPHP, as all the servers are registered at init.

lib/request-processor/main.go
```diff
@@ -38,49 +45,85 @@ func initializeServer(server *ServerData) {
...
-	log.Debugf("Init data: %s", initJson)
-	log.Debugf("Started with token: \"AIK_RUNTIME_***%s\"", utils.AnonymizeToken(globals.CurrentToken))
+	log.Debugf(instance, "Init data: %s", initJson)
+	log.Debugf(instance, "Started with token: \"AIK_RUNTIME_***%s\"", utils.AnonymizeToken(instance.GetCurrentToken()))
 
 	if globals.EnvironmentConfig.PlatformName != "cli" {
-		grpc.Init()
-		server := globals.GetCurrentServer()
+		server := instance.GetCurrentServer()
+
```
tudor-timcu:
```suggestion
if globals.EnvironmentConfig.PlatformName != "cli" {
		server := instance.GetCurrentServer()
        server.ServerInitMutex.Lock()
        defer server.ServerInitMutex.Unlock()

        if server.ServerInitialized {
            server.ServerInitMutex.Unlock()
            return true
        }
        grpc.Init()

		if server != nil {
			initializeServer(server)
			grpc.StartCloudConfigRoutine()
			server.ServerInitialized = true
		}
	}

	if !zen_internals.Init() {
```

lib/request-processor/aikido_types/handle.go
```diff
@@ -25,4 +24,5 @@ type RequestShutdownParams struct {
 	QueryParsed         map[string]interface{}
 	IsWebScanner        bool
 	ShouldDiscoverRoute bool
+	IsIpBypassed        bool
```
PopoviciMarian: `IsIpBypassed` is no longer needed in `RequestShutdownParams`

lib/request-processor/handle_request_metadata.go
```diff
@@ -21,42 +21,53 @@ func OnRequestShutdownReporting(params RequestShutdownParams) {
...
+			RouteParsed:    context.GetParsedRoute(instance),
+			StatusCode:     context.GetStatusCode(instance),
+			User:           context.GetUserId(instance),
+			UserAgent:      context.GetUserAgent(instance),
+			IP:             context.GetIp(instance),
+			Url:            context.GetUrl(instance),
+			RateLimitGroup: context.GetRateLimitGroup(instance),
+			RateLimited:    context.IsEndpointRateLimited(instance),
+			QueryParsed:    context.GetQueryParsed(instance),
+			IsIpBypassed:   context.IsIpBypassed(instance),
```
PopoviciMarian: `- IsIpBypassed`

lib/request-processor/handle_request_metadata.go
```diff
@@ -21,42 +21,53 @@ func OnRequestShutdownReporting(params RequestShutdownParams) {
 		return
 	}
 
-	log.Info("[RSHUTDOWN] Got request metadata: ", params.Method, " ", params.Route, " ", params.StatusCode)
-
-	params.IsWebScanner = webscanner.IsWebScanner(params.Method, params.Route, params.QueryParsed)
+	log.Info(nil, "[RSHUTDOWN] Got request metadata: ", params.Method, " ", params.Route, " ", params.StatusCode)
 
+	if !params.IsIpBypassed {
```
PopoviciMarian: this `if` should be deleted

lib/agent/globals/globals.go
```diff
@@ -10,6 +10,7 @@ var Machine MachineData
 var Servers = make(map[ServerKey]*ServerData)
 var PastDeletedServers = make(map[ServerKey]bool)
 var ServersMutex sync.RWMutex
+var RegisteredServersMutex sync.RWMutex
```
tudor-timcu: this should be deleted, we should use only one mutex to sync servers -> ServersMutex

lib/agent/server_utils/server.go
```diff
@@ -29,9 +29,15 @@ func storeConfig(server *ServerData, req *protos.Config) {
 }
 
 func Register(serverKey ServerKey, requestProcessorPID int32, req *protos.Config) {
+	globals.RegisteredServersMutex.RLock()
```
tudor-timcu:
```suggestion
	globals.ServersMutex.Lock()
    defer globals.ServersMutex.Unlock()
```

lib/php-extension/RequestProcessor.cpp
```diff
@@ -222,42 +273,50 @@ void RequestProcessor::LoadConfig(const std::string& previousToken, const std::s
...
-    this->requestProcessorConfigUpdateFn(GoCreateString(initJson));
+    std::string initJson = requestProcessor.GetInitData(currentToken);
+    requestProcessor.requestProcessorConfigUpdateFn(this->requestProcessorInstance, GoCreateString(initJson));
 }
 
-void RequestProcessor::LoadConfigFromEnvironment() {
-    std::string previousToken = AIKIDO_GLOBAL(token);
+void RequestProcessorInstance::LoadConfigFromEnvironment() {
+    auto& globalToken = AIKIDO_GLOBAL(token);
+    std::string previousToken = globalToken;
```
tudor-timcu: can we revert this to look as before:

```suggestion
    std::string previousToken = AIKIDO_GLOBAL(token);
    LoadEnvironment();
    std::string currentToken = AIKIDO_GLOBAL(token);
    LoadConfig(previousToken, currentToken);
```

lib/request-processor/config/config.go
```diff
@@ -2,22 +2,24 @@ package config
...
 	"main/utils"
 	"os"
 )
 
-func UpdateToken(token string) bool {
-	if token == globals.CurrentToken {
-		log.Debugf("Token is the same as previous one, skipping config reload...")
+func UpdateToken(instance *instance.RequestProcessorInstance, token string) bool {
+	server := globals.GetServer(token)
+	if token == instance.GetCurrentToken() {
```
tudor-timcu: something look odd here: server should be initialize after this check and i think it should be checked for nil. also the log message does not match the if clause

tudor-timcu (approved):
Congratz again!

PR #424, opened 2026-05-17
Build for ARM 64 alpine

### LINKED ISSUE #167: Alpine Linux Support
Is there any plan to add a build for Alpine Linux?  My deployed production image is built from `php:8.4-fpm-alpine` since it ends up being about 500MB smaller than using `php:8.4-fpm` and I'd love to be able to include this in my version without having to switch over to Debian based images.

tylersatre:
This fixes #167 

I decided to take a look at this using codex and was able to get it working for an ARM64 Alpine container when I installed it and set it up.  I know there's work to be done to wire up releases, but since it's a small change in the code I wanted to at least PR that to see if it could be brought in.

### DIFF
--- a/lib/request-processor/vulnerabilities/zen-internals/zen_internals.go
+++ b/lib/request-processor/vulnerabilities/zen-internals/zen_internals.go
@@ -25,6 +25,7 @@ import (
 	"main/globals"
 	"main/log"
 	"main/utils"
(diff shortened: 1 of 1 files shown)


kidk:
Thanks @tylersatre apologies for delay in our reply. We will take a look

PR #440, opened 2026-08-07
Optimize path traversal check

hansott:


teta2k:
I benchmarked the follow-up against the original PR #440 implementation using the same request-level benchmark (`-benchtime=500ms -count=5 -benchmem`; medians shown):

| Workload | PR #440 | Follow-up | Result |
|---|---:|---:|---|
| 5 MB body, 235 reads | 24.31 ms | 22.17 ms | Equivalent |
| 50k plain fields, 235 reads | 80.93 ms | 81.67 ms | Equivalent |
| 50k relative paths, 235 reads | 808.50 ms | 79.18 ms | ~10.2x faster |
| Early body detection | 1.954 µs | 1.957 µs | Equivalent |

The follow-up improves two things:

- It only retains inputs that can reach a path-traversal detection branch (`../` / `..\`, or an absolute path) instead of treating every value containing `/` or `\` as a candidate. This prevents large bodies containing harmless URLs or relative paths from being rescanned on every filesystem sink.
- It no longer preallocates the candidate map with `len(full)`. On the 50k-field cases this reduced allocation by about 5.25 MB per request.

The cache remains request-scoped and source-lazy. Safe benchmarks assert no detection, and the early-attack benchmark confirms that a body match still avoids parsing a very large later headers source.

Using `*map` for consistency with the other request caches adds one 8-byte allocation per request, with no material timing impact.

lib/request-processor/vulnerabilities/path-traversal/checkContextForPathTraversal.go
```diff
@@ -36,6 +37,12 @@ func CheckContextForPathTraversal(instance *instance.RequestProcessorInstance, f
 	return nil
 }
 
+func isPathTraversalCandidate(input string) bool {
+	sanitizedInput := SanitizePath(helpers.TrimInvisible(input))
+	userInput := helpers.ExtractResourceOrOriginal(sanitizedInput)
+	return len(sanitizedInput) > 1 && (containsUnsafePathParts(userInput) || filepath.IsAbs(userInput))
+}
```
hansott, line 60: perhaps we should move this closer to `GetPathTraversalCandidates`? where it's actually used? now it's passed in a callback style but we can just call the function I guess?

teta2k: Refactored a bit

teta2k (approved):

PR #446, opened 2026-08-13
Report attack wave stats

timokoessler:


lib/agent/grpc/request.go
```diff
@@ -32,6 +32,13 @@ func storeAttackStats(server *ServerData, req *protos.AttackDetected) {
 	}
 }
 
+func storeAttackWaveStats(server *ServerData) {
+	server.StatsData.StatsMutex.Lock()
+	defer server.StatsData.StatsMutex.Unlock()
+
+	server.StatsData.AttackWaves += 1
```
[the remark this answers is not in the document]
timokoessler: We don't support Attack wave blocking yet, so this is on purpose (mirrored from firewall-node).

lib/agent/grpc/request.go
```diff
@@ -32,6 +32,13 @@ func storeAttackStats(server *ServerData, req *protos.AttackDetected) {
 	}
 }
 
+func storeAttackWaveStats(server *ServerData) {
```
hansott, line 35: incrementAttackWaveStats?

timokoessler: The existing methods all start with `store`, e.g. `storeAttackStats`, but happy to change.

lib/agent/cloud/event_heartbeat.go
```diff
@@ -129,6 +133,8 @@ func GetStatsAndClear(server *ServerData) Stats {
 	server.StatsData.RequestsAborted = 0
```
teta2k, line 133: Unrelated, but perhaps we should also clear RequestsRateLimited now that we have the unit test?

teta2k (approved):

PR #462, opened 2026-08-25
Fix missing PHP extension logs when disk logging is disabled

teta2k:
## Why

`AIKIDO_DISK_LOGS` is disabled by default. In that mode, the PHP extension logger selected `stdout` as its destination but then checked whether its disk-log file was open. Because no disk file is created when disk logging is disabled, `Log::Write` returned without writing anything.

As a result, every PHP extension warning and error was silent by default, including failures to load or initialize the request processor, failures to start the agent process, and exceptions inside protection handlers. These are the messages users need when the extension is not working.

## What changes

| Configuration | PHP extension behavior |
| --- | --- |
| Disk logging disabled, INFO or DEBUG | No output |
| Disk logging disabled, WARN or ERROR | Write to `stderr` |
| Disk logging enabled | Continue writing configured log levels to the per-process disk log |

The configured `AIKIDO_LOG_LEVEL` threshold still applies. `stderr` is used instead of `stdout` so extension diagnostics do not become normal PHP or CLI output and remain visible in container, FPM, and web-server logs.

## Why the hook messages become INFO

Making warnings visible would also expose existing messages for hook targets that are not installed, such as PostgreSQL, MySQLi, PDO, or cURL functions and classes. Missing optional APIs are expected and do not mean the agent is malfunctioning, so those messages are changed from WARN to INFO. They remain available in verbose disk logs, while genuine hook conflicts and initialization failures remain WARN or ERROR.

## Documentation

Updates the troubleshooting guide to explain where to find Zen logs by default, how to enable additional logs with `AIKIDO_DISK_LOGS`, how `AIKIDO_LOG_LEVEL` affects their contents, and when the service or container must be restarted.

## Validation

Adds PHPT regression coverage proving that an extension error is emitted when disk logging is disabled.

Validated with a PHP 8.5 extension build, all 6 targeted `aikido_ops` tests, and a broad CLI run where 78 passed, 1 skipped, and only the 11 PostgreSQL cases could not run because the local image lacks `pg_connect`. All 238 GitHub CI jobs passed.

### DIFF
--- a/tests/cli/aikido_ops/test_extension_logs_without_disk_logs.phpt
+++ b/tests/cli/aikido_ops/test_extension_logs_without_disk_logs.phpt
@@ -1,0 +1,16 @@
+--TEST--
+PHP extension logs errors when disk logs are disabled
+
+--ENV--
+AIKIDO_LOG_LEVEL=ERROR
+AIKIDO_DISK_LOGS=0
+
+--FILE--
+<?php
+
+\aikido\set_token("");
+
+?>
+
+--EXPECTREGEX--
+.*\[AIKIDO\]\[ERROR\]\[\d+\]\[\d+\]\[[0-9:]+\] set_token: token is null!.*
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -17,7 +17,21 @@ Test from the same environment where your app runs and follow the instructions o
 
 ## Check logs for errors
 
-`cat /var/log/aikido-*/*`
+The agent writes its main log to `/var/log/aikido-<version>/aikido-agent-*-main.log`:
+
+```sh
+cat /var/log/aikido-*/aikido-agent-*-main.log
+```
+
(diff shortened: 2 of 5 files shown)


tests/cli/aikido_ops/test_extension_logs_without_disk_logs.phpt
```diff
@@ -0,0 +1,16 @@
+--TEST--
+PHP extension logs errors when disk logs are disabled
```
hansott, line 2: it requires `catch_workers_output = yes` to be enabled in `/etc/php/<version>/fpm/pool.d/www.conf`, we enable it for our tests but the default is false (see https://www.php.net/manual/en/install.fpm.configuration.php)

should we mention that somewhere in the docs?

teta2k: Nice catch. Added in the documentation

hansott (approved):

PR #475, opened 2026-09-03
Fix FrankenPHP worker request initialization

teta2k:
## Problem

FrankenPHP worker mode has two lifecycles:

1. PHP runs normal `RINIT`/`RSHUTDOWN` around the long-lived worker script.
2. Every HTTP request handled inside that script needs its own Aikido init/shutdown.

The old shared init code returned early whenever `FRANKENPHP_WORKER=1`. That skipped the worker boot as intended, but it also skipped every manual `\aikido\worker_rinit()` call. Requests were served without an Aikido request context, so worker traffic was not protected.

## Why the tests were green

The generated worker script was outside Caddy's document root, so FrankenPHP served the tests in classic mode. The handler also called nonexistent `\aikido\worker_request_init()` instead of the exported `\aikido\worker_rinit()`. Because the handler never ran, that error stayed hidden.

## Why the real worker test exposed heap corruption

Once the harness entered worker mode, two abnormal paths needed explicit handling:

- A blocking exception skips code after the application include unless shutdown is in `finally`.
- `exit()`/`die()` bypasses PHP `finally`; PHP invokes native `RSHUTDOWN` while FrankenPHP unwinds that request.

The initial fix returned early from native `RSHUTDOWN` for every worker. On `exit()`, Aikido's AST/request state therefore outlived the PHP request heap. Reusing that worker thread later produced `zend_mm_heap corrupted`.

## Fix

- Detect and cache worker mode during native `RINIT`, skipping only the worker-script boot initialization.
- Use `\aikido\worker_rinit()` for each handled HTTP request.
- Run `\aikido\worker_rshutdown()` in `finally` for normal returns and exceptions.
- Use native `RSHUTDOWN` as the `exit()`/`die()` fallback.
- Use one common shutdown entry point. The existing request shutdown keeps its disabled fast-path and AST cleanup ordering; `RequestProcessorInstance::RequestShutdown()` now owns the initialization guard and is safe to call more than once.
- Run the generated worker as `index.php` inside its Caddy root, reset the test app working directory on every request, and keep the IMDS fixture request-local by inlining its one-use helper.

Worker mode is read from the environment only during native startup and then cached. Manual request init/shutdown only read the boolean.

## Verification

- Reproduced the original real-worker behavior: requests were served without protection.
- Reproduced the exit failure: FrankenPHP terminated with `zend_mm_heap corrupted` and missing shutdowns.
- Focused rate-limit regression reports all 115 requests, including 10 `exit()`-based rate-limit responses, with no heap error.
- Focused worker blocking tests pass for SSRF and path traversal, including runtime blocking-mode changes.
- FrankenPHP classic-mode control test passes.
- CI for the preceding implementation: https://github.com/AikidoSec/firewall-php/actions/runs/33812774994


lib/php-extension/RequestProcessor.cpp
```diff
@@ -212,7 +212,9 @@ bool RequestProcessorInstance::RequestInit() {
     std::string sapiName = sapi_module.name;
 
     if (sapiName == "frankenphp") {
-        if (GetEnvBoolWithAllGetters("FRANKENPHP_WORKER", false)) {
+        // FRANKENPHP_WORKER stays set for real worker requests, so only the
+        // first automatic RINIT is warm-up; worker_rinit() must continue below.
```
hansott: cleanup?

teta2k: Refactored for better clarity

hansott (approved):

PR #486, opened 2026-09-10
Update Go to 1.26.8 and gRPC to 1.83.2

teta2k:
Published Zen PHP v1.5.19 binaries contain gRPC 1.83.1, which is flagged by [GHSA-2v4p-qf9q-27wj / CVE-2026-84445](https://github.com/grpc/grpc-go/security/advisories/GHSA-2v4p-qf9q-27wj). Update both Go modules to the patched gRPC 1.83.2 and bump the official Go pins to 1.26.8.

The advisory's xDS server path is unreachable in these binaries: the agent uses `grpc.NewServer` over a Unix socket, and the request processor is a client. Source inspection, production package graphs, and inspection of both architectures' published binaries found no xDS server usage.

Replace build-time `go get` and `go mod tidy` with `go mod download` and `go mod verify`, so builds use the reviewed module versions. The old commands explain why v1.5.19 shipped gRPC 1.83.1 while its source declared 1.83.0. Update existing Go download checksums and use Go's default compiler selection based on the module requirements. Alpine retains its exact custom compiler source commit; its official bootstrap pin becomes 1.26.8.

Define the zero-field `Empty` response in the shared IPC schema and use its generated Go type in the agent. The responses retain their empty protobuf encoding, while generation no longer depends on the system's `google/protobuf/empty.proto`. Both modules use the modern `google.golang.org/protobuf` runtime without declaring the legacy `github.com/golang/protobuf` module. The existing Ubuntu 20.04 CI image and plain `protoc` commands are sufficient.

Validated locally on Linux amd64:

- With Ubuntu 20.04's original `protoc 3.6.1` and Go 1.26.8, both modules pass their full test suites and executable/shared-library builds under `-mod=readonly`; `go.mod` and `go.sum` remain unchanged.
- The shared library passes the `dlopen` smoke test on glibc 2.31. Both rebuilt binaries report gRPC 1.83.2 and contain no legacy Go protobuf module.
- Module verification, Go formatting, workflow validation, and modified shell-script syntax checks pass. Full actionlint reports the same 65 existing shellcheck warnings as main, with no new diagnostics.

Native arm64 builds, Alpine validation of the final IPC schema, and full PHP/package integration remain for CI. The Alpine compiler image was not rebuilt with the new bootstrap pin. This PR prepares the dependency fix; stable release publication is separate.


lib/agent/go.mod
```diff
@@ -1,10 +1,11 @@
 module main
 
-go 1.26.6
+go 1.26.8
 
 require (
+	github.com/golang/protobuf v1.5.4
```
timokoessler: Wasn't this actually replaced by https://pkg.go.dev/google.golang.org/protobuf which is also used below?

teta2k: Solved

timokoessler (approved):

